PR #7, opened 2022-02-01
SDI-33 Basic usage of SQS queue

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
```diff
@@ -0,0 +1,20 @@
...
+  private companion object {
+    val log: Logger = LoggerFactory.getLogger(this::class.java)
+  }
+
+  fun migrateVisits() {
+    queueMigrationService.sendMessage(MIGRATE_VISITS)
+  }
+
+  fun migrateVisitsByPage() = log.info("Will calculate visit pages to migrate")
+}
```
andymarke, line 20: so clearly we are doing diddly squat right now

src/main/resources/application-localstack.yml
```diff
@@ -2,5 +2,5 @@ hmpps.sqs:
   provider: localstack
   queues:
     migration:
-      queueName: migration_queue
-      dlqName: migration_dlq
+      queueName: ${random.uuid}
+      dlqName: ${random.uuid}
```
andymarke, line 6: avoid issues of multiple Listeners listening to same queue during unit testing

mikehalmamoj (approved):

PR #11, opened 2022-02-04
SDI-33 Divide batch into page chunks

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/data/MigrationContext.kt
```diff
@@ -3,6 +3,8 @@ package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.data
 import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
 
-class MigrationContext<T>(val migrationId: String, val estimatedCount: Long, val filter: T)
+class MigrationContext<T>(val migrationId: String, val estimatedCount: Long, val body: T) {
```
andymarke, line 6: renamed to body, more flexible then "filter"

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/listeners/MigrationMessageListener.kt
```diff
@@ -28,6 +29,7 @@ class MigrationMessageListener(
     val migrationMessage: MigrationMessage<*> = message.fromJson()
     when (migrationMessage.type) {
       MIGRATE_VISITS -> visitsMigrationService.migrateVisitsByPage(context(message.fromJson()))
+      MIGRATE_VISITS_BY_PAGE -> visitsMigrationService.migrateVisitsForPage(context(message.fromJson()))
```
petergphillips, line 32: this reads a bit odd `MIGRATE_VISITS` calls `migrateVisitsByPage` but `MIGRATE_VISITS_BY_PAGE` calls something else

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/listeners/MigrationMessageListener.kt
andymarke: Yes, they are sort of state transitions I suppose, 
MIGRATE_VISITS -> MIGRATE_VISITS_BY_PAGE and each function is the transition to the next stage, hence it looks odd

PR #14, opened 2022-02-07
SDI-33 Add nomis visit mapping dependency

andymarke:
1. Added nomis visit mapping dependency 
2. Added localstack test containers


petergphillips (approved):
LGTM

PR #15, opened 2022-02-07
SDI-33 Add Visit Mapping service and check visit has not been migrated already

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
```diff
@@ -69,9 +70,15 @@ class VisitsMigrationService(
     )
   }.forEach { queueService.sendMessage(MIGRATE_VISIT, it) }
 
-  fun migrateVisit(context: MigrationContext<VisitId>) {
-    log.info("Migrating visit {}", context.body.visitId)
-  }
+  fun migrateVisit(context: MigrationContext<VisitId>) =
+    visitMappingService.findNomisVisitMapping(context.body.visitId)
+      ?.run {
```
andymarke, line 75: Can't think of the scenario this might happen except message is sent again, but it would be bad if we allowed this so best to log error

petergphillips (approved):
LGTM

PR #19, opened 2022-02-08
SDI-33 Retrieve room mapping

andymarke:
I have outstanding question which is what do we do if we find no room

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitMappingServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitMappingServiceTest.kt
@@ -124,4 +124,63 @@ internal class VisitMappingServiceTest {
       )
     }
   }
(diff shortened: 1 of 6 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitMappingServiceTest.kt
```diff
@@ -124,4 +124,63 @@ internal class VisitMappingServiceTest {
...
+        )
+      )
+
+      assertThat(visitMappingService.findRoomMapping("HB7SOC", "HEI")).isNull()
+    }
+
+    @Test
+    internal fun `will return the mapping when found`() {
+      visitMappingApi.stubFor(
+        get(urlPathMatching("/prison/.+?/room/nomisRoomId/.+?")).willReturn(
```
andymarke, line 149: guessing on the url and response for now

lbennett-moj (approved):

PR #20, opened 2022-02-09
SDI-33 Actually map the data and push to VSIP

andymarke:


lbennett-moj (approved):
LGTM, with change to action on missing room mapping

PR #22, opened 2022-02-10
SDI-33 Final step:create the vsip nomis mapping

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
```diff
@@ -76,36 +76,44 @@ class VisitsMigrationService(
...
+          val vsipVisit = visitsService.createVisit(
+            mapNomisVisit(nomisVisit, this)
+          )
+          visitMappingService.createNomisVisitMapping(
+            nomisVisitId = nomisVisit.visitId,
+            vsipVisitId = vsipVisit.visitId,
+            migrationId = context.migrationId
+          )
+        } ?: run {
+          log.error("Could not find room mapping for visit ${nomisVisit.visitId} with prison ${nomisVisit.prisonId} and agency internal location ${nomisVisit.agencyInternalLocation?.description}")
```
andymarke, line 98: confirmed with VSIP Team, no mapping should result in error

lbennett-moj: I thought it would just log the error and not throw an exception - we don't want it on the DLQ do we?

andymarke: I was thinking of sticking on DLQ so we have a record, and VSIP can then decide to either fix forward or discard all DLQ messages

lbennett-moj (approved):

PR #24, opened 2022-02-11
SDI-33 Fix issue where mapping failure was retried

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
```diff
@@ -147,7 +148,7 @@ class VisitsMigrationService(
         "nomisVisitId" to nomisVisitId.toString()
       ),
       null
-    ).also { throw e }
+    )
```
andymarke, line 151: we swallow exception and manual procedure to fix this comes into operation

lbennett-moj (approved):

PR #28, opened 2022-02-15
SDI-40 Draft of integrating more migration information into /info endpoint

andymarke:
- use /info as a "at a glance" progress of a migration

TODO
- plug into real mapping endpoints


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitMappingService.kt
```diff
@@ -45,8 +47,14 @@ class VisitMappingService(@Qualifier("visitMappingApiWebClient") private val web
       }
       .block()
   }
+
+  fun findLatestMigration(): LatestMigration? = LatestMigration()
```
andymarke, line 51: call real endpoints when available

petergphillips (approved):
LGTM

PR #29, opened 2022-02-15
SDI-40 Upgrade localstack to version that has arm64 version

andymarke:


docker-compose-localstack.yaml
```diff
@@ -1,15 +1,14 @@
...
-    image: localstack/localstack:0.12.10
+    image: localstack/localstack:0.14.0
     networks:
       - hmpps
     container_name: localstack
     ports:
       - "4566-4597:4566-4597"
     environment:
       - SERVICES=sqs
-      - DATA_DIR=/tmp/localstack/data
```
andymarke, line 12: not sure why this was here, but it hung localstack

mikehalmamoj (approved):

PR #34, opened 2022-02-28
SDI-71: 📝 Copy date time fix and security improvements

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationFilter.kt
```diff
@@ -14,7 +14,7 @@ data class VisitsMigrationFilter(
     description = "List of visit types to migrate",
     example = "SCON",
     allowableValues = ["SCON", "OFFI"],
-    defaultValue = "SCON"
+    defaultValue = "[\"SCON\"]"
```
petergphillips, line 17: docs test failed due to this not be an list

Mjwillis (approved):
LGTM

PR #38, opened 2022-03-07
SDI-98 Pass through ignoreMissingRoom to nomis api to filter out VSIP  visits during testing

andymarke:
There is an edge case that almost certainly would only affect testing migration / synchronisation 

- Sync a VSIP visit with NOMIS
- Delete all mappings to force a new full migration 
- All VSIP visits will fail migration because they have no room data
This adds a flag to the migration process to ignore all visits that have no rooms (essentially VSIP visits)

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
@@ -19,6 +19,7 @@ class NomisApiService(@Qualifier("nomisApiWebClient") private val webClient: Web
     visitTypes: List<String>,
     fromDateTime: LocalDateTime?,
     toDateTime: LocalDateTime?,
+    ignoreMissingRoom: Boolean,
     pageNumber: Long,
     pageSize: Long,
   ): PageImpl<VisitId> =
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
```diff
@@ -19,6 +19,7 @@ class NomisApiService(@Qualifier("nomisApiWebClient") private val webClient: Web
     visitTypes: List<String>,
     fromDateTime: LocalDateTime?,
     toDateTime: LocalDateTime?,
+    ignoreMissingRoom: Boolean,
```
lbennett-moj, line 22: would it be better as nullable? It will default to false in the api if parameter not provided

lbennett-moj (dismissed):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
andymarke: I did consider that, but went for the default of false which functionally will work the same

lbennett-moj (approved):

PR #40, opened 2022-03-09
SDI-104 Added logic to calculate when migration is finished

andymarke:
The change adds an additional telemetry event when we think the migration is finished.

This estimates that a migration is finished when there are no message on the migration queue (though there might be messages in process)

Since SQS does not guarantee the count is correct, this change tries several times over a 10 second period to check if there are no more messages on the queue leaving a very small possibility we will mark the migration as completed when we are very close to the end - the impact of this operational is negligible


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/MigrationQueueService.kt
```diff
@@ -31,5 +37,13 @@ class MigrationQueueService(
     )
   }
 
+  // given counts are approximations there is only a probable chance this returns the correct result
+  fun isItProbableThatThereAreStillMessagesToBeProcessed(): Boolean =
+    migrationSqsClient.countMessagesOnQueue(migrationQueueUrl) > 0
+
```
andymarke, line 43: decided to ignore messages in progress since worse case scenario is it just about to finish (and this message itself would be in-progress)

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
```diff
@@ -77,6 +79,13 @@ class VisitsMigrationService(
...
         queueService.sendMessage(MIGRATE_VISITS_BY_PAGE, it)
       }
+    queueService.sendMessage(
+      MIGRATE_VISITS_STATUS_CHECK,
+      MigrationContext(
+        context = context,
+        body = VisitMigrationStatusCheck()
+      )
+    )
   }
```
andymarke, line 89: this is sent after all pages have been chunked

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
```diff
@@ -132,6 +141,43 @@ class VisitsMigrationService(
...
+    */
+    if (queueService.isItProbableThatThereAreStillMessagesToBeProcessed()) {
+      queueService.sendMessage(
+        MIGRATE_VISITS_STATUS_CHECK,
+        MigrationContext(
+          context = context,
+          body = VisitMigrationStatusCheck()
+        ),
+        delaySeconds = 10
+      )
```
andymarke, line 157: check every ten seconds

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
```diff
@@ -183,6 +229,9 @@ class VisitsMigrationService(
   }
 }
 
+private fun <T> MigrationContext<T>.durationMinutes(): Long =
+  Duration.between(LocalDateTime.parse(this.migrationId), LocalDateTime.now()).toMinutes()
+
```
andymarke, line 234: for now relay on the migrationId being the date time migration started - if we ever made the migrationId a GUUID then obviously this would have to change

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/integration/visits/VisitsMigrationIntTest.kt
```diff
@@ -138,6 +140,14 @@ class VisitsMigrationIntTest : SqsIntegrationTestBase() {
 
       verify(telemetryClient).trackEvent(eq("nomis-migration-visits-started"), any(), isNull())
       verify(telemetryClient, times(26)).trackEvent(eq("nomis-migration-visit-migrated"), any(), isNull())
+
+      await.atMost(Duration.ofSeconds(21)) untilAsserted {
```
andymarke, line 144: it should never take more than 21 seconds since we check every 10 seconds followed by 10 checks 1 second apart

lbennett-moj (approved):
LGTM

PR #42, opened 2022-03-10
SDI-104 Add integration with postgres database (no read/write yet)

andymarke:
Adds integration to Postgres but yet to actually use it for anything
1) Should work in Cloud Platform
2) health check works locally with just postgres and localstack running
3) health check works locally when everything is working in localstack


docker-compose.yml
```diff
@@ -119,12 +120,23 @@ services:
       - hmpps
     container_name: hpfnm-mapping-db
     restart: unless-stopped
-    ports:
-      - "5432:5432"
     environment:
```
andymarke, line 123: doesn't need to be exposed outside of docker network

lbennett-moj (approved):

PR #46, opened 2022-03-14
SDI-104 Add counts to history record

andymarke:
- Reads DLQ to record number of failures
- Read visit migration table to read number of successes

The DLQ is not 100% reliable since there could DLQ messages on queue already, but will mention this in Run Book when done

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/MigrationQueueService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/MigrationQueueService.kt
@@ -20,6 +20,8 @@ class MigrationQueueService(
   private val migrationQueue by lazy { hmppsQueueService.findByQueueId("migration") as HmppsQueue }
   private val migrationSqsClient by lazy { migrationQueue.sqsClient }
   private val migrationQueueUrl by lazy { migrationQueue.queueUrl }
+  private val migrationDLQSqsClient by lazy { migrationQueue.sqsDlqClient!! }
+  private val migrationDLQUrl by lazy { migrationQueue.dlqUrl!! }
 
   fun sendMessage(message: Messages, context: MigrationContext<*>, delaySeconds: Int = 0) {
     val result =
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/MigrationQueueService.kt
```diff
@@ -20,6 +20,8 @@ class MigrationQueueService(
   private val migrationQueue by lazy { hmppsQueueService.findByQueueId("migration") as HmppsQueue }
   private val migrationSqsClient by lazy { migrationQueue.sqsClient }
   private val migrationQueueUrl by lazy { migrationQueue.queueUrl }
+  private val migrationDLQSqsClient by lazy { migrationQueue.sqsDlqClient!! }
+  private val migrationDLQUrl by lazy { migrationQueue.dlqUrl!! }
```
andymarke, line 25: these are nullable in library - but shouldn't be null for this service - hence `!!`

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/integration/visits/VisitsMigrationIntTest.kt
```diff
@@ -159,19 +165,25 @@ class VisitsMigrationIntTest : SqsIntegrationTestBase() {
...
-        .expectStatus().isOk
-        .expectBody()
-        .jsonPath("$.size()").isEqualTo(1)
-        .jsonPath("$[0].migrationId").isNotEmpty
-        .jsonPath("$[0].whenStarted").isNotEmpty
-        .jsonPath("$[0].whenEnded").isNotEmpty
-        .jsonPath("$[0].estimatedRecordCount").isEqualTo(26)
-        .jsonPath("$[0].migrationType").isEqualTo("VISITS")
-        .jsonPath("$[0].status").isEqualTo("COMPLETED")
+      await untilAsserted {
```
andymarke, line 168: wrap the whole thing in a retry else occasionally fails since migration is in the processing of ending, but hasn't quite  written the row yet

lbennett-moj (approved):

PR #49, opened 2022-03-15
SDI-104 Add Visits specific history endpoint

andymarke:
- Generic `history` endpoint now protected by admin role
- Visits specific version of endpoint 

This is required for 2 reasons:

- Visit migration users should only see visit migrations
- Has ability to understand the filter, so can easily and reliable filter by prison

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitMigrationResource.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitMigrationResource.kt
@@ -1,23 +1,34 @@
 package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.visits
 
 import io.swagger.v3.oas.annotations.Operation
+import io.swagger.v3.oas.annotations.Parameter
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitMigrationResource.kt
```diff
@@ -51,4 +62,59 @@ class VisitMigrationResource(private val visitsMigrationService: VisitsMigration
...
+      example = "HEI",
+    ) @RequestParam prisonId: String? = null,
+  ) = migrationHistoryService.findAll(
+    HistoryFilter(
+      migrationTypes = listOf(MigrationType.VISITS.name),
+      fromDateTime = fromDateTime,
+      toDateTime = toDateTime,
+      includeOnlyFailures = includeOnlyFailures,
+      filterContains = prisonId?.let { """"prisonIds":["$it"]""" }
+    )
```
andymarke, line 118: can only query migrations that are done on a second prison, which is how the migration will be rolled out

petergphillips (approved):
LGTM

PR #98, opened 2022-06-14
SDI-227 Work around for rate limited purge

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/MigrationQueueService.kt
```diff
@@ -48,13 +56,32 @@ class MigrationQueueService(
...
-    )
+    }.onFailure {
+      deleteAllMessages()
+    }
+  }
+
+  private fun deleteAllMessages() {
+    kotlin.runCatching {
+      val messageCount = migrationSqsClient.countMessagesOnQueue(migrationQueueUrl)
+      repeat(messageCount) {
```
lbennett-moj, line 76: is this count always going to be accurate?

andymarke: Probably not but this is called multiple times whe. Cancelling

PR #99, opened 2022-06-14
SDI 227 cancel purge delay

andymarke:
A bit experimental; but attempt to purge queue every second for half a minute to remove messages being added by other threads

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
@@ -331,9 +331,7 @@ class VisitsMigrationService(
       null
     )
     migrationHistoryService.recordMigrationCancelledRequested(migrationId)
-    queueService.purgeAllMessages()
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitsMigrationService.kt
```diff
@@ -331,9 +331,7 @@ class VisitsMigrationService(
       null
     )
     migrationHistoryService.recordMigrationCancelledRequested(migrationId)
-    queueService.purgeAllMessages()
```
andymarke, line 334: delay sending message until the initial 30 seconds of purging is complete

src/test/resources/application-test.yml
```diff
@@ -33,3 +33,6 @@ spring:
 #  level:
 #   io.r2dbc.postgresql.QUERY: DEBUG # for queries
 #    io.r2dbc.postgresql.PARAM: INFO # for parameters
+cancel:
+  queue:
+    purge-time-attempts: 2
```
andymarke, line 38: so we can have sensibly timed integration test

PR #100, opened 2022-06-14
SDI-227 Made global scope so doesn't suspend

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/MigrationQueueService.kt
```diff
@@ -96,21 +96,19 @@ class MigrationQueueService(
     }
   }
 
-  suspend fun purgeAllMessagesNowAndAgainInTheNearFuture(migrationContext: MigrationContext<VisitMigrationStatusCheck>) {
+  fun purgeAllMessagesNowAndAgainInTheNearFuture(migrationContext: MigrationContext<VisitMigrationStatusCheck>) {
     purgeAllMessages()
     // so that MIGRATE_VISITS_BY_PAGE messages that are currently generating large numbers of messages
     // have their messages immediately purged, keep purging messages every second for around 10 seconds
-    coroutineScope {
```
andymarke, line 103: this was using the scope for this `suspend` function and hence blocking the rest call, this should now just just a global thread, effectively fire and forget

PR #116, opened 2022-07-07
SDI-264: 🩺 Add database health checks

petergphillips:


helm_deploy/hmpps-prisoner-from-nomis-migration/values.yaml
```diff
@@ -72,3 +72,4 @@ generic-service:
 
 generic-prometheus-alerts:
   targetApplication: hmpps-prisoner-from-nomis-migration
+  alertSeverity: syscon
```
petergphillips, line 75: moved common value up a level

PR #127, opened 2022-07-21
sdi-262 enabling auditable username to be read in a reactive service,…

lbennett-moj:
… plus conversion to reactive with the exception of a messaging listener that must remain synchronous


andymarke (approved):
LGTM

PR #138, opened 2022-08-22
SDI-355 Very basic flow that just reads IDs and does nothing

andymarke:
* Flow just calls the IDs API and outputs custom events
* Refactor some of the queue stuff so it was queue agnostic, queue ID now held on context

TODO
* Complete migration flow
* Unit tests for NOMIS API
* Maybe Unit tests for service (questioning the value of those)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/data/MigrationContext.kt
```diff
@@ -1,10 +1,16 @@
 package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.data
 
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.MigrationType
 import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
 
-class MigrationContext<T>(val migrationId: String, val estimatedCount: Long, val body: T) {
-  constructor(context: MigrationContext<*>, body: T) : this(context.migrationId, context.estimatedCount, body)
```
andymarke, line 7: MigrationType is "the" enum representing what type of migration. Need so we can make the queue id generic

lbennett-moj (approved):
LGTM

PR #144, opened 2022-08-30
SDI-355 incentives migration

andymarke:
Add call to retrieve IEP from NOMIS 
TODO - migrate it to incentives!


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
```diff
@@ -131,6 +131,16 @@ class NomisApiService(@Qualifier("nomisApiWebClient") private val webClient: Web
       .retrieve()
       .bodyToMono(NomisIncentive::class.java)
       .awaitSingle()
+
```
andymarke, line 134: blocking version since this is called from a listener

mikehalmamoj (approved):

PR #167, opened 2022-09-26
sdi-388 ignore events originating in VSIP

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/VisitSynchronisationService.kt
```diff
@@ -24,29 +25,36 @@ class VisitSynchronisationService(
...
+          log.debug("Ignoring visit cancellation event for ${this@run} as no nomis visit mapping found")
+        }
+      } else {
+        log.debug("Ignoring VSIP-originated visit cancellation event for ${this@run}")
       }
     }
   }
+
+  private fun NomisVisit.nomisCancellation() =
+    this.modifyUserId != "PRISONER_MANAGER_API" // doesn't originate from VSIP
```
andymarke, line 59: I wonder if we should make this an environment property or anything? I think for now hardcode it but might be worth thinking if there is a better way

andymarke (approved):
LGTM

PR #197, opened 2022-12-02
sdi-474 create time edge case

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
```diff
@@ -228,7 +228,10 @@ data class NomisIncentive(
      We will use the seconds from the 'when created' timestamp to avoid multiple IEP records created in the same minute being out of order.
    */
   private fun getTransformedIncentiveDateTime(): LocalDateTime =
-    iepDateTime.withSecond(whenCreated.second)
+    if (iepDateTime.minute != whenCreated.minute) {
```
lbennett-moj, line 231: Gone for != rather than less as this is bound to happen at midnight 😆

andymarke (approved):
LGTM

PR #216, opened 2023-01-27
SDI-593: 🔇 More selective muting of spring scheduling in application insights

petergphillips:


src/main/resources/application-dev.yml
```diff
@@ -8,7 +8,8 @@ api.base.url:
   visits: http://localhost:8083
   visit.mapping: http://localhost:8084
   sentencing: http://localhost:8085
-
+  mapping: http://localhost:8083
```
petergphillips, line 11: couldn't start the application in dev profile without setting these two

PR #260, opened 2023-03-15
Sdit 644 visits

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visits/MigrationVisitsMessageListener.kt
```diff
@@ -1,60 +1,48 @@
...
-    }.thenAccept { }
+  fun onVisitMessage(message: String, rawMessage: Message): CompletableFuture<Void>? {
+    return onMessage(message, rawMessage)
   }
 
-  private inline fun <reified T> String.fromJson(): T =
-    objectMapper.readValue(this, object : TypeReference<T>() {})
+  override fun parseContextFilter(json: String): MigrationMessage<*, VisitsMigrationFilter> {
+    return objectMapper.readValue(json)
+  }
```
mikehalmamoj, line 35: I don't understand star projections very well... but wouldn't you know the type at this point?

andymarke: No, because it is a class so the type gets lost at runtime unlike the reified inline functions

andymarke (approved):
LGTM

PR #273, opened 2023-04-11
SDI-702 Allow DPS delete to return 404 without error

andymarke:
This allows this theoretical edge case to work:

* Adjustment created in DPS
* Adjustment delete in DPS
* Update service deletes NOMIS record 
* Triggers messages this service so this service also attempts to delete adjustment that has already been deleted
* Update service deletes Mapping
* Call to DPS service returns 404 since adjustment is already deleted

So treat DELETE in DPS as idempotent

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/sentencing/SentencingSynchronisationIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/sentencing/SentencingSynchronisationIntTest.kt
@@ -532,9 +532,9 @@ class SentencingSynchronisationIntTest : SqsIntegrationTestBase() {
   @DisplayName("SENTENCE_ADJUSTMENT_DELETED")
   inner class SentenceAdjustmentDeleted {
     @Nested
-    inner class WhenDeletedByDPS {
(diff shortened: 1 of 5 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/sentencing/SentencingSynchronisationIntTest.kt
```diff
@@ -636,6 +636,60 @@ class SentencingSynchronisationIntTest : SqsIntegrationTestBase() {
...
+      @DisplayName("When deleted by DPS but mapping not deleted yet")
+      inner class WhenAdjustmentHasBeenDeleted {
+        @BeforeEach
+        fun setUp() {
+          mappingApi.stubGetNomisSentencingAdjustment(
+            adjustmentCategory = "SENTENCE",
+            nomisAdjustmentId = NOMIS_ADJUSTMENT_ID,
+            adjustmentId = ADJUSTMENT_ID,
+          )
+          sentencingApi.stubDeleteSentencingAdjustmentForSynchronisationNotFound(ADJUSTMENT_ID)
```
andymarke, line 650: really just shows the 404 will be ignored

mikehalmamoj (approved):

PR #289, opened 2023-05-10

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/appointments/AppointmentsService.kt
```diff
@@ -1,39 +1,59 @@
...
+  val startDate: LocalDate?,
+
+  @field:NotNull(message = "Start time must be supplied")
+  @JsonFormat(pattern = "HH:mm")
+  val startTime: LocalTime?,
+
+  @JsonFormat(pattern = "HH:mm")
+  val endTime: LocalTime?,
+
+  val comment: String = "",
```
steverendell, line 58: I copied this over as-is from the activities api. Maybe not-null fields should not be nullable though!

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/appointments/AppointmentsService.kt
```diff
@@ -1,39 +1,59 @@
...
-  val status: String,
+data class CreateAppointmentResponse(
+  val id: Long,
 )
 
-data class CreateAppointmentResponse(
-  val appointmentInstanceId: Long,
+data class AppointmentMigrateRequest(
+
+  @field:NotEmpty(message = "Prison code must be supplied")
```
mikehalmamoj, line 30: How come all of these are nullable types but also have validation for NotEmpty or NotNull?

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/appointments/AppointmentsService.kt
steverendell: Will do in next iteration

PR #290, opened 2023-05-12

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
```diff
@@ -216,17 +216,26 @@ data class AppointmentResponse(
...
+  val createdDate: LocalDateTime,
+  val createdBy: String,
+  val modifiedDate: LocalDateTime? = null,
+  val modifiedBy: String? = null,
 ) {
   fun toAppointment() = AppointmentMigrateRequest(
     bookingId = bookingId,
     prisonerNumber = offenderNo,
-    prisonCode = prisonId,
+    prisonCode = prisonId!!, // never null in existing nomis data for event_type = 'APP' (as at 11/5/2023)
```
mikehalmamoj, line 227: In that case I'd be tempted to make the entity field non-nullable. If you get a null it's going to blow up at some point anyway but at least this would communicate that it shouldn't be null (even if the Nomis column is nullable).

steverendell: ok will do

mikehalmamoj (approved):

PR #294, opened 2023-05-16

steverendell:
Fix appointments endpoint

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -69,7 +69,7 @@ java {
 }
 
 tasks {
-  withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile> {
+  withType<KotlinCompile> {
     //  dependsOn("buildSentencingApiModel")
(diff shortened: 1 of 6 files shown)


build.gradle.kts
```diff
@@ -95,7 +95,7 @@ tasks.register("buildActivityApiModel", GenerateTask::class) {
   apiPackage.set("uk.gov.justice.digital.hmpps.prisonerfromnomismigration.activities.api")
   configOptions.set(
     mapOf(
-      "dateLibrary" to "java8-localdatetime",
+      "dateLibrary" to "string",
```
mikehalmamoj, line 98: Interesting. I was expecting the localdatetime one but that always felt wrong because the description is "legacy only"!

steverendell: Well. ... This is a good one, the problem here is that there are no @JsonFormats in the generated dtos so the json created for the dates in the request dto is completely wrong. So reverted to conversion to string manually

mikehalmamoj (approved):

PR #318, opened 2023-08-04
SDIT-941 improved telemetry

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/adjudications/AdjudicationsMigrationService.kt
```diff
@@ -79,17 +79,21 @@ class AdjudicationsMigrationService(
...
         )
 
         telemetryClient.trackEvent(
           "adjudications-migration-entity-migrated",
           mapOf(
             "adjudicationNumber" to adjudicationNumber.toString(),
+            "chargeSequence" to chargeSequence.toString(),
+            "chargeNumber" to chargeNumber,
             "migrationId" to context.migrationId,
           ),
```
andymarke, line 98: interestingly we don't have  test for telemetry since we check the bulk telemetry (start, complete etc).
I'm thinking we need an integration test that better describes the behaviour of a single item being migrated with some better asserts

mikehalmamoj (approved):

PR #328, opened 2023-08-11
SDIT-928 Add other parties to DPS DTO

andymarke:
Prisoners can either be

- suspects (so in this scenario other suspects involved in the incident)
- victims
- witnesses

Staff can either be

- victim
- witness

Reporting Officer (oddly along side the main reporting officer)
Control and restraint (so I guess used in the incident)
Other
So give we have a single “witness” list with types of OFFICER, STAFF, OTHER_PERSON, VICTIM, PRISONER I think the best I can do it is as follows:

- All staff witnesses - become STAFF
- All prisoners witnesses - become OTHER_PERSON (obviously there are very few of these)
- All victims (staff or prisoners) - become VICTIM
- All other prisoners (e.g other suspects) become PRISONER
- All other staff (the last 3 on that list) becomes OTHER_PERSON

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/adjudications/AdjudicationTransformationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/adjudications/AdjudicationTransformationTest.kt
@@ -1,10 +1,12 @@
 package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.adjudications
 
 import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.DisplayName
 import org.junit.jupiter.api.Nested
 import org.junit.jupiter.api.Test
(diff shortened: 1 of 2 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/adjudications/AdjudicationTransformationTest.kt
```diff
@@ -300,12 +312,283 @@ class AdjudicationTransformationTest {
...
+            staffId = 2,
+            firstName = "KWEKU",
+            lastName = "KOFI",
+          ),
+        ),
+      )
+      val dpsAdjudication = nomisAdjudication.toAdjudication()
+
+      assertThat(dpsAdjudication.witnesses).hasSize(2)
+      assertThat(dpsAdjudication.witnesses).contains(
```
mikehalmamoj, line 350: You could do `containsExactlyInAnyOrder`.

mikehalmamoj (approved):

PR #334, opened 2023-08-14
SDIT-928 Additional mapping of usernames and sanction sequence

andymarke:
As well as additional mapping the the adjudication response rom NOMIS now only holds outcomes for the specific charge


lbennett-moj (approved):
lgtm

PR #338, opened 2023-08-14
SDIT-996 Add minimal config for non-associations api

karenmillermoj:


src/test/resources/application-test.yml
```diff
@@ -26,6 +27,9 @@ appointments:
 adjudications:
   page:
     size: 10
+non-associations:
+  page:
+    size: 10
```
karenmillermoj, line 32: Not needed yet - but copy and paste and will be needed later

andymarke (approved):
LGTM

PR #339, opened 2023-08-14
SDIT-928 Calculate award days when supplied in months

andymarke:
DPS always records in days but NOMIS sometimes (rarely) uses months; so 2 months wage deduction. The best we can do is convert this back into days based on when punishment started (so February is a good month to be punished)

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/sentencing/SentencingMigrationIntTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/sentencing/SentencingMigrationIntTest.kt
@@ -9,6 +9,7 @@ import org.awaitility.kotlin.untilAsserted
 import org.awaitility.kotlin.untilCallTo
 import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.BeforeEach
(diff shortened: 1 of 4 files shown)


mikehalmamoj (approved):

src/test/resources/application-test.yml
```diff
@@ -47,8 +47,8 @@ spring:
 #    io.r2dbc.postgresql.PARAM: INFO # for parameters
 cancel:
   queue:
-    purge-frequency-time: 20ms
-    purge-total-time: 1s
+    purge-frequency-time: 100ms
+    purge-total-time: 5s
```
andymarke, line 52: revert back to original timing - test was failing sometimes

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/sentencing/SentencingMigrationIntTest.kt
```diff
@@ -223,6 +224,7 @@ class SentencingMigrationIntTest : SqsIntegrationTestBase() {
     }
 
     @Test
+    @Disabled
```
andymarke, line 227: another flaky test waiting for Steve to fix

mikehalmamoj (approved):

PR #340, opened 2023-08-15
Fix Activities URLs in preprod and prod.

mikehalmamoj:


iterative-it (approved):
LGTM. Checked and verified those are the correct base urls for the Activities API

PR #363, opened 2023-08-22
SDIT-996 Add sync for non-association

karenmillermoj:


mikehalmamoj (approved):
Looks good - I was just wondering if it also needs a couple of negative tests. What happens when any of the API calls fail?

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
```diff
@@ -230,6 +231,13 @@ class NomisApiService(@Qualifier("nomisApiWebClient") private val webClient: Web
       .retrieve()
       .bodyToMono(typeReference<RestResponsePage<FindActiveAllocationIdsResponse>>())
       .awaitSingle()
+
+  suspend fun getNonAssociation(offenderNo: String, nsOffenderNo: String): NonAssociationResponse =
+    webClient.get()
+      .uri("/non-associations/offender/{offenderNo}/ns-offender/{nsOffenderNo}", offenderNo, nsOffenderNo)
+      .retrieve()
+      .bodyToMono(NonAssociationResponse::class.java)
```
andymarke, line 239: you can just do awaitBody instead of these 2 (and can drop the class name)

andymarke (approved):
LGTM

PR #364, opened 2023-08-22
SDIT-928 Use correct consecutive adjudication number

andymarke:
Consecutive awards can be associated with a different adjudication, this sets the correct parent number


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/wiremock/NomisApiMockServer.kt
```diff
@@ -1024,9 +1024,11 @@ fun adjudicationResponse(
                                 "statusDate": "2023-08-07",
                                 "sanctionDays": 2,
                                 "sanctionMonths": 1,
-                                "chargeSequence": 1
+                                "chargeSequence": 1,
+                                "adjudicationNumber": ${adjudicationNumber - 1}
```
andymarke, line 1028: random default value

lbennett-moj (approved):

PR #366, opened 2023-08-22
SDIT-996 Ensure paired non-association-events are treated as one

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsPrisonOffenderEventListener.kt
```diff
@@ -42,8 +42,8 @@ class NonAssociationsPrisonOffenderEventListener(
                   (sqsMessage.Message.fromJson()),
                 )
 
-              "NON_ASSOCIATION_DETAIL_DELETED" -> log.debug(
```
karenmillermoj, line 45: This just corrects the deleted event name

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsSynchronisationService.kt
```diff
@@ -23,12 +23,19 @@ class NonAssociationsSynchronisationService(
...
+      telemetryClient.trackEvent(
+        "non-association-synchronisation-duplicate-skipped",
+        event.toTelemetryProperties(),
+      )
       return
     }
 
     nomisApiService.getNonAssociation(event.offenderIdDisplay, event.nsOffenderIdDisplay)
-      ?.also {
+      .also {
```
karenmillermoj, line 38: A new PR will mean this will change shortly (so makes no difference at the mo)

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsSynchronisationService.kt
```diff
@@ -38,6 +45,9 @@ class NonAssociationsSynchronisationService(
   }
 }
 
+// Two non-association events occur for each non-association relationship - use the offenderIds to uniquely identify each pair
+private fun NonAssociationsOffenderEvent.isDuplicate(): Boolean = offenderIdDisplay > nsOffenderIdDisplay
```
karenmillermoj, line 49: Not sure if this is a good name - possibly isNonAssociationIdentifier or isNonAssociationPrimary? all ideas welcome

mikehalmamoj: I think I prefer isNonAssociationPrimary because of the various drawbacks of using the term Duplicate in this scenario. 
I do wonder if the best thing is to deduplicate these events in prison-offender-events by using the alphabetical order trick. What do you think @andymarke?

mikehalmamoj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsSynchronisationService.kt
andymarke: Hmm, not sure since "other" clients might want both events - in fact I think we raise to domain events which might be what clients want, so I think doing it here works slightly better.

PR #372, opened 2023-08-24
SDIT-928 tactical short term change to "ignore" 4xx errors

andymarke:
While DPS is understanding their mapping issues they will throw a 4xx error. For these, when we are in the special "reporting" mode, we ignore the error, but gather some stats


steverendell (approved):
lgtm

PR #376, opened 2023-08-31
SDIT-984 Add telemetry when an entity fails to migrate

mikehalmamoj:
Otherwise it's hard to find which entity it was


andymarke (approved):
lgtm

PR #392, opened 2023-09-08
SDIT-1085 Temporary hack to prevent null sanction seq from causing error

andymarke:
Currently DPS does not return a mapping for some punishments it can not process, this cause the entire migration record to fail..

While this is WIP in DPS this change ignores the missing mapping and just warnings.

Code hack can be revered once DPS is fixed but right now it gets in the way of analysis


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/adjudications/AdjudicationsMigrationService.kt
```diff
@@ -357,7 +359,13 @@ class AdjudicationsMigrationService(
...
       )
     } ?: emptyList(),
-    punishments = mapping.punishmentMappings?.map {
+    punishments = mapping.punishmentMappings?.filter {
+      (it.sanctionSeq != null).also { hasSanction ->
+        if (!hasSanction) {
+          log.warn("Ignoring mapping for punishment ${it.punishmentId} for booking ${it.bookingId}. DPS has no sanctionSeq as this is a WIP")
+        }
+      }
+    }?.map {
```
andymarke, line 368: just log and filter out where sanctionSeq is null - which really means DPS can't create a punishment for now (but will once the analysis has been done)

lbennett-moj (approved):

PR #394, opened 2023-09-11
SDIT-1092: ⬆️ Upgrade to latest to fix vuln.

petergphillips:


build.gradle.kts
```diff
@@ -10,24 +10,12 @@ plugins {
   id("org.openapi.generator") version "7.0.0"
 }
 
-dependencyCheck {
-  suppressionFiles.add("reactive-suppressions.xml")
```
petergphillips, line 14: turns out that the gradle OWASP check passes without any of these suppressions

andymarke (approved):
lgtm

PR #400, opened 2023-09-13
SDIT-997 Add non-associations migration

karenmillermoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/integration/SqsIntegrationTestBase.kt
```diff
@@ -108,15 +112,15 @@ class SqsIntegrationTestBase : TestBase() {
 
   private val allQueues by lazy {
     listOf(
```
karenmillermoj, line 114: re-ordered, added nonAssoocationsMigrationQueue and removed duplicate

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationResource.kt
```diff
@@ -0,0 +1,232 @@
...
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.MigrationType
+import java.time.LocalDateTime
+
+@RestController
+@RequestMapping("/migrate", produces = [MediaType.APPLICATION_JSON_VALUE])
+class NonAssociationsMigrationResource(
+  private val nonAssociationsMigrationService: NonAssociationsMigrationService,
+  private val migrationHistoryService: MigrationHistoryService,
+) {
+  @PreAuthorize("hasRole('ROLE_MIGRATE_NON_ASSOCIATIONS')")
```
petergphillips, line 36: if the same role restrictions are applicable to all under here then might be better to put the `PreAuthorize` up a level and then if a new method is added the role restriction will then apply automatically

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationResource.kt
```diff
@@ -0,0 +1,232 @@
...
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.config.ErrorResponse
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.persistence.repository.MigrationHistory
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.HistoryFilter
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.InProgressMigration
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.MigrationHistoryService
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.MigrationType
+import java.time.LocalDateTime
+
+@RestController
+@RequestMapping("/migrate", produces = [MediaType.APPLICATION_JSON_VALUE])
```
petergphillips, line 31: would it be better for the request mapping here to include `/non-associations` too - all endpoints look to be at that level

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/wiremock/NonAssociationsApiMockServer.kt
```diff
@@ -147,6 +148,9 @@ class NonAssociationsApiMockServer : WireMockServer(WIREMOCK_PORT) {
 
   fun createNonAssociationSynchronisationCount() =
     findAll(putRequestedFor(urlMatching("/sync/upsert"))).count()
+
+  fun createNonAssociationMigrationCount() =
+    findAll(postRequestedFor(urlMatching("/migrate"))).count()
```
petergphillips, line 153: I would have expected this to include the `/non-associations` path too?

petergphillips (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/wiremock/NonAssociationsApiMockServer.kt
karenmillermoj: That doesn't work - it is only matching on the /migrate

PR #407, opened 2023-09-19
SDIT-997 Ensure check for non-primary association in migration

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationService.kt
```diff
@@ -65,9 +66,22 @@ class NonAssociationsMigrationService(
...
   override suspend fun migrateNomisEntity(context: MigrationContext<NonAssociationIdResponse>) {
-    log.info("attempting to migrate ${context.body}")
-    val firstOffenderNo = context.body.offenderNo1
-    val secondOffenderNo = context.body.offenderNo2
+    val nonAssociation = context.body
+    log.info("attempting to migrate $nonAssociation")
+    val firstOffenderNo = nonAssociation.offenderNo1
+    val secondOffenderNo = nonAssociation.offenderNo2
+
+    if (nonAssociation.isNotPrimaryNonAssociation()) {
```
karenmillermoj, line 74: Bomb straight out if this is not a non-association we're interested in.

Mjwillis (dismissed):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationService.kt
petergphillips: might look better to do a 
```kotlin
with (context.body) {
   log.info("attempting to migrate $this")
   ...
```
and you won't need to define any of the 3 variables

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationService.kt
```diff
@@ -65,9 +66,22 @@ class NonAssociationsMigrationService(
...
+      telemetryClient.trackEvent(
+        "non-association-migration-non-primary-skipped",
+        mapOf(
+          "firstOffenderNo" to firstOffenderNo,
+          "secondOffenderNo" to secondOffenderNo,
+          "migrationId" to context.migrationId,
+        ),
+      )
+      return
```
petergphillips, line 83: `return` not doing anything?

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationService.kt
```diff
@@ -180,5 +194,8 @@ class NonAssociationsMigrationService(
   }
 }
 
+// Two non-association events occur for each non-association relationship - use the offenderIds to uniquely identify each pair
+fun NonAssociationIdResponse.isNotPrimaryNonAssociation(): Boolean = offenderNo1 > offenderNo2
```
petergphillips, line 198: not sure why this is a `>` rather than a `!=`?

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationServiceTest.kt
```diff
@@ -978,6 +978,28 @@ internal class NonAssociationsMigrationServiceTest {
       )
     }
 
+    @Test
+    internal fun `will not migrate non-primary associations`(): Unit = runBlocking {
```
petergphillips, line 982: think we're trying to use `runTest` instead of `runBlocking` as it skips any delays (but not sure there will be any difference here)

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationServiceTest.kt
```diff
@@ -978,6 +978,28 @@ internal class NonAssociationsMigrationServiceTest {
...
+      )
+
+      verify(telemetryClient).trackEvent(
+        eq("non-association-migration-non-primary-skipped"),
+        check {
+          assertThat(it["firstOffenderNo"]).isEqualTo("D5678EF")
+          assertThat(it["secondOffenderNo"]).isEqualTo("A1234BC")
+          assertThat(it["migrationId"]).isNotNull
+        },
+        eq(null),
```
petergphillips, line 999: better to use `isNull()`

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationService.kt
karenmillermoj: We get 2 events for offender pair - but to ensure we only use one event (to uniquely identify the pair) we decided to only use the offender pair where the first is smaller than the last

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationService.kt
karenmillermoj: return just bombs out - it follows the pattern done in other service methods where we don't process the event if we're not interested

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationService.kt
petergphillips: I assumed that the primary was the one where the first was the same as the second.

Might be better for the comment to say something like
```
// Two non-association events occur for each non-association relationship.
// We call the primary association the one where the first offender number is greater (string comparison) than the second.
```

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationService.kt
petergphillips: sorry, I hadn't realised there was more code after the return 🙁

petergphillips (dismissed):
LGTM

petergphillips (approved):
LGTM

PR #417, opened 2023-09-28
SDIT-997 Reorder migrated offender numbers to ensure firstOffenderNo is smallest

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/nonassociations/NonAssociationsMigrationService.kt
```diff
@@ -70,12 +69,15 @@ class NonAssociationsMigrationService(
     with(context.body) {
       log.info("attempting to migrate $this")
 
+      val smallestOffenderNo = if (offenderNo1 < offenderNo2) offenderNo1 else offenderNo2
+      val largestOffenderNo = if (offenderNo1 < offenderNo2) offenderNo2 else offenderNo1
+
```
karenmillermoj, line 74: Re-order to ensure we deal with the offenders in the order we want.  Smallest should be the firstOffender

mikehalmamoj (approved):

PR #418, opened 2023-09-28
SDIT-1112 Document manual procedure for creating mappings

andymarke:


README.md
```diff
@@ -183,9 +183,82 @@ curl --location --request POST 'https://nomis-sync-prisoner-mapping.hmpps.servic
     "label": "2022-02-14T09:58:45"
 }'
 ```
+#### Adjudications
+ 
+##### DSP Migration 500 error resolution
 
+A 500 error returned from DPS service `POST /reported-adjudications/migrate` typically means we have previously migrated an adjudication and the migration service is trying to migrate the same record again.
+This can happen when the `POST /reported-adjudications/migrate` was successfully by a network issue meant the migration service never received the response. The migration service will retry the migration and the DPS service will return a 500 error since it detects a duplicate record.
```
steverendell, line 191: is there a word missing around "was successfully by a network issue"

PR #445, opened 2023-11-15
SDIT-1247 Try to create mapping with DPS returns a 406

andymarke:
This wiring in the mapping creation service.

The scenario is as follows

- Try migrate adjudication
- DPS POST fails with timeout but does successfully migrate item
- No mapping is created due to POST timeout
- Retry message
- DPS POST now fails with a 406 (meaning adjudication already migrated)
- Ignore 406 but instead programmatically create mapping ourselves

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/adjudications/AdjudicationsMigrationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/adjudications/AdjudicationsMigrationService.kt
@@ -253,6 +253,7 @@ class AdjudicationsMigrationService(
   auditService: AuditService,
   private val adjudicationsMappingService: AdjudicationsMappingService,
   private val adjudicationsService: AdjudicationsService,
+  private val mappingCreator: AdjudicationMappingCreator,
   @Value("\${adjudications.page.size:1000}") pageSize: Long,
   @Value("\${complete-check.delay-seconds}") completeCheckDelaySeconds: Int,
   @Value("\${complete-check.count}") completeCheckCount: Int,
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/adjudications/AdjudicationsMigrationService.kt
```diff
@@ -310,20 +311,37 @@ class AdjudicationsMigrationService(
...
+                "chargeNumber" to mapping.chargeNumberMapping.chargeNumber,
+                "offenderNo" to offenderNo,
+                "migrationId" to context.migrationId,
+              ),
+              null,
+            )
+          } ?: run {
+            // unlikely to get here since we checked the mapping just above
+            telemetryClient.trackEvent(
+              "adjudications-migration-entity-already-migrated",
```
mikehalmamoj, line 334: I don't get it - looks like this gets run if the createAdjudication call returns a 406 - how does 406 = already migrated?

lbennett-moj (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/adjudications/AdjudicationsMigrationService.kt
andymarke: DPS defines it so - they return 406 if that get a JPA duplicate key error

mikehalmamoj: Not mod security then 🤯

PR #450, opened 2023-11-20
SDIT-1233 DPS rejects exclusions if there is no schedule for that time

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/activities/AllocationsMigrationService.kt
```diff
@@ -170,14 +172,15 @@ private fun GetAllocationResponse.toAllocationMigrateRequest(activityId: Long, s
...
+    exclusions = exclusions.toDpsExclusions(scheduleRules),
   )
 
-fun List<AllocationExclusion>.toDpsExclusions(): List<Slot> =
+fun List<AllocationExclusion>.toDpsExclusions(scheduleRules: List<ScheduleRulesResponse>): List<Slot> =
   listOf("AM", "PM", "ED").map { timeSlot ->
     this
       .filter { exclusion -> exclusion.slot == null || exclusion.slot.value == timeSlot }
       .map { exclusion -> exclusion.findDay() }
+      .filter { day -> scheduleRules.includes(day, timeSlot) }
```
mikehalmamoj, line 183: NOMIS allows exclusions that don't match the schedule but DPS doesn't - so don't send any exclusions with no schedule.

lbennett-moj (approved):

PR #459, opened 2023-12-20
SDIT-1347 Pass on errors received from NOMIS api

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
```diff
@@ -216,6 +216,12 @@ class NomisApiService(@Qualifier("nomisApiWebClient") private val webClient: Web
             Mono.error(BadRequestException(errorResponse.userMessage ?: "Received a 400 calling /activities/ids"))
           }
       },)
+      .onStatus({ it == HttpStatus.NOT_FOUND }, { clientResponse ->
+        clientResponse.bodyToMono(ErrorResponse::class.java)
+          .flatMap { errorResponse ->
+            Mono.error(NotFoundException(errorResponse.userMessage ?: "Received a 404 calling /activities/ids"))
+          }
+      },)
```
mikehalmamoj, line 224: Previously receiving a not found from nomis-prisoner-api due to invalid prison ID was sending a 400 back to the UI - as I'm putting in some better error handling in the UI we need to pass on the correct error.

PR #460, opened 2023-12-20
SDIT-1347 Handle null exception cause

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
```diff
@@ -210,13 +211,11 @@ class NomisApiService(@Qualifier("nomisApiWebClient") private val webClient: Web
           .build()
       }
       .retrieve()
-      .onStatus({ it == HttpStatus.BAD_REQUEST }, { clientResponse ->
-        clientResponse.bodyToMono(ErrorResponse::class.java)
-          .flatMap { errorResponse ->
-            Mono.error(BadRequestException(errorResponse.userMessage ?: "Received a 400 calling /activities/ids"))
-          }
-      },)
```
mikehalmamoj, line 218: There was a weird bug caused by this where the `ExceptionHandler` had a null `cause` and we ended up sending a null `userMessage` back to the caller. Don't get the same problem with `onErrorResume`.

andymarke (approved):
lgtm

PR #461, opened 2023-12-21
SDIT-1347 Fix null error message

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/config/ExceptionHandler.kt
```diff
@@ -82,7 +82,7 @@ class HmppsPrisonerFromNomisMigrationExceptionHandler {
         .body(
           ErrorResponse(
             status = BAD_REQUEST,
-            userMessage = "Bad Request: ${e.cause?.message}",
```
mikehalmamoj, line 85: This was copied from the ValidationException where it makes sense - doesn't make sense here as the cause is usually null.

steverendell (approved):

PR #476, opened 2024-01-24
SDIT-1351 Remove unused endpoints

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/activities/ActivitiesMigrationResource.kt
```diff
@@ -210,58 +208,6 @@ class ActivitiesMigrationResource(
   )
   suspend fun getActiveMigrationDetails() = migrationHistoryService.getActiveMigrationDetails(MigrationType.ACTIVITIES)
 
-  @PreAuthorize("hasRole('ROLE_MIGRATE_ACTIVITIES')")
-  @GetMapping("/ids")
```
mikehalmamoj, line 214: All of the removed endpoints used to be called by the migration dashboard, but that's been changed to call nomis prisoner API directly so these are no longer used.

andymarke (approved):
lgtm

PR #483, opened 2024-02-12
SDIT-1487 Add incidents sync

karenmillermoj:


steverendell (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/incidents/IncidentsSynchronisationService.kt
```diff
@@ -66,6 +67,35 @@ class IncidentsSynchronisationService(
     }
   }
 
+  suspend fun synchroniseIncidentDelete(event: IncidentsOffenderEvent) {
+    // Should never happen
+    if (event.auditModuleName == "DPS_SYNCHRONISATION") {
```
andymarke, line 72: BTW I suspect you won't get the the auditModule correct for a delete, since the NOMIS record is gone so the auditModule would be the previous update for this, so you might end up calling delete in DPS anyway and ignoring a 404

andymarke (approved):
LGTM

PR #486, opened 2024-02-12
SDIT-1370 Add prison Id to incident migrate request

karenmillermoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/wiremock/NomisApiMockServer.kt
```diff
@@ -1387,15 +1391,3 @@ private fun suspendedAllocationsResponse(): String =
       }
     ]
   """
-
```
karenmillermoj, line 1390: Was used by non-associations - no longer needed

petergphillips (approved):
LGTM

PR #487, opened 2024-02-13
SDIT-1375 Reduce Incidents Listener threads to 8

karenmillermoj:
Connection pool size is 10 and default thread listener is 10 - when any other endpoint is called that requires a connection (e.g. healthcheck), it can use more connections than in the pool - hence 500 errors. This is a particular issue during early stages of incidents migration due to the get IDs endpoint being so CPU intensive.
This is the same change made to the AdjudicationsMigrationMessageListener.


petergphillips (approved):
LGTM

PR #492, opened 2024-02-19
SDIT-1505 Integration with alerts DPS APIs

andymarke:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsDpsApiMockServer.kt
```diff
@@ -32,6 +40,25 @@ class AlertsDpsApiMockServer : WireMockServer(WIREMOCK_PORT) {
...
+      alertSeq = 2,
+      alertUuid = UUID.randomUUID(),
+      status = CREATED,
+    ),
+  ) {
+    stubFor(
+      post("/alerts")
+        .willReturn(
+          aResponse()
+            .withStatus(200)
```
mikehalmamoj, line 55: If their endpoint is an upsert then a 200 would make sense - or is it a 201 for create and 200 for update?

iterative-it: Yes it does use 201 to indicate that the request created an alert in the new service and 200 when an existing alert was updated instead

andymarke: thanks

andymarke: BTW @iterative-it I couldn't see the response code in the swagger (not that it matters so long as it is in the 2xx range)

PR #494, opened 2024-02-20
SDIT-1505 Handle simple create alert scenario

andymarke:
* Assumes DPS API uses NOMIS model - we know this will change
* TODO - need prisoner number
* TODO - handle mapping errors
* TODO - handle mapping already exists

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/Transformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/Transformer.kt
@@ -1,0 +1,44 @@
+package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.alerts
+
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.alerts.model.NomisAlert
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomissync.model.AlertResponse
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+// The initial contract with DPS was to send the entire NOMIS alert, hence the DTO name. This is likely to change
+// but for now copy as if we are sending all this data
+@Suppress("ktlint:standard:discouraged-comment-location")
+fun AlertResponse.toDPsAlert() = NomisAlert(
+  alertDate = this.date,
+  offenderBookId = this.bookingId,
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationIntTest.kt
```diff
@@ -1,42 +1,173 @@
...
+
+  @Autowired
+  private lateinit var alertsMappingApiMockServer: AlertsMappingApiMockServer
 
   @Nested
   @DisplayName("ALERT-INSERTED")
   inner class AlertInserted {
+
     @Nested
-    @DisplayName("Check queue config")
```
andymarke, line 23: no longer needs this placeholder test now we are testing "for real"

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/Transformer.kt
```diff
@@ -0,0 +1,44 @@
+package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.alerts
+
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.alerts.model.NomisAlert
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomissync.model.AlertResponse
+import java.time.LocalDateTime
+import java.time.format.DateTimeFormatter
+
+// The initial contract with DPS was to send the entire NOMIS alert, hence the DTO name. This is likely to change
+// but for now copy as if we are sending all this data
+@Suppress("ktlint:standard:discouraged-comment-location")
```
andymarke, line 10: I'm going to ask to see how we can remove this rule, not convinced it helps anyone

mikehalmamoj: I've read the docs and think the rule is correct. The problem is that this
```offenderNo = "A1234TT", // TODO - return in NOMIS API``` when read by a parser looks like the comment applies to the line below (because it comes after the `,`). Hence it's always better to put the comment above the line in a parameter list, e.g.
```
 // TODO - return in NOMIS API
offenderNo = "A1234TT",
```

mikehalmamoj (approved):

PR #501, opened 2024-02-22
SDIT-1519 Implement alert synchronization logic

andymarke:
This commit introduces alert synchronization logic in the service. If an ALERT-UPDATED event is received for an alert that has already been synchronized (checked by the audit module name), the event is ignored and a telemetry event is tracked. If the event is for a new alert, depending on whether the mapping exists or not, a failure or success telemetry event is tracked and appropriate actions are taken.


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
```diff
@@ -65,7 +65,28 @@ class AlertsSynchronisationService(
   }
 
   suspend fun nomisAlertUpdated(event: AlertUpdatedEvent) {
-    log.debug("TODO: handle {}", event)
+    // hello
```
lbennett-moj, line 68: obviously a crucial comment

lbennett-moj (approved):

PR #530, opened 2024-03-25
SDIT-1614 Alert migration plumbing

andymarke:
When DPS sees a duplicate from NOMIS it will return a 409 - so the intention is to log and ignore these


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsMigrationMessageListener.kt
```diff
@@ -19,7 +20,7 @@ import java.util.concurrent.CompletableFuture
 class AlertsMigrationMessageListener(
   objectMapper: ObjectMapper,
   alertsMigrationService: AlertsMigrationService,
-) : MigrationMessageListener<AlertsMigrationFilter, NomisAlertId, AlertResponse, AlertMappingDto>(
```
andymarke, line 22: this DTO has now been defined in NOMIS API

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsDpsApiMockServer.kt
```diff
@@ -48,10 +50,9 @@ class AlertsDpsApiMockServer : WireMockServer(WIREMOCK_PORT) {
       prisonNumber = "A1234AA",
       alertCode = AlertCodeSummary(
         alertTypeCode = "A",
```
andymarke, line 52: changes to the DPS API

mikehalmamoj (approved):

PR #534, opened 2024-03-26
SDIT-1620 From-nomis migration and sync of location deactivation data

steverendell:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/locations/LocationsMigrationService.kt
```diff
@@ -232,9 +232,21 @@ fun toUpsertSyncRequest(nomisLocationResponse: LocationResponse, parentId: Strin
 
     createDate = nomisLocationResponse.createDatetime,
     // lastModifiedDate - no value available as it changes with occupancy
-    deactivatedDate = nomisLocationResponse.deactivateDate,
```
andymarke, line 235: do these have valves even when they the location is active?

andymarke (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/locations/LocationsMigrationService.kt
steverendell: They get cleared in P-Nomis when you set the location active, but potentially that may not always happen. Will get trapped by the reconciliation.

PR #535, opened 2024-03-26
SDIT-1620 From-nomis migration and sync of location deactivation data

steverendell:
Always set deactivation reason if not active


andymarke (approved):
lgtm

PR #556, opened 2024-04-09
SDIT-1619 Locations reconciliation - improve migration

steverendell:
Fix problem with duplicate history items by just updating history in the main location migeration dto


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/locations/LocationsSynchronisationService.kt
```diff
@@ -146,3 +152,270 @@ private fun LocationsOffenderEvent.toTelemetryProperties(
 ) + (dpsLocationId?.let { mapOf("dpsLocationId" to it) } ?: emptyMap()) + (
   mappingFailed?.let { mapOf("mapping" to "initial-failure") } ?: emptyMap()
   )
+
```
steverendell, line 155: This section is basically a copy of the migration mappings but just using the separate NomisSyncLocationRequest type

Mjwillis (approved):
LGTM

PR #558, opened 2024-04-11
SDIT-1658 First pass at skeletion for migrating alerts by prisoner

andymarke:
This is feature switched via alerts.migration.type with test by default using old method and deployed service using skeleton of new.

This allows us to hedge our bets depending on what we find when only migrating some alerts from previous bookings. If we discover this is not acceptable we can easily revert back to old method.

This version only track telemetry event showing how many previous alerts will be made active by this approach.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsByPrisonerMigrationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsByPrisonerMigrationService.kt
@@ -1,0 +1,95 @@
+package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.alerts
+
+import com.microsoft.applicationinsights.TelemetryClient
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import org.springframework.beans.factory.annotation.Value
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsByPrisonerMigrationService.kt
```diff
@@ -0,0 +1,95 @@
...
+  override suspend fun migrateNomisEntity(context: MigrationContext<PrisonerId>) {
+    log.info("attempting to migrate ${context.body}")
+    val latestNomisBookingId = context.body.bookingId
+    val offenderNo = context.body.offenderNo
+    val status = context.body.status
+
+    val nomisAlerts = alertsNomisService.getAlertsToMigrate(offenderNo)
+    val activeAlertsFromPreviousBookings = nomisAlerts.previousBookingsAlerts.filter { it.isActive }
+    alertsDpsService.migrateAlerts(nomisAlerts.latestBookingAlerts.map { it.toDPSMigratedAlert(context.body.offenderNo) })?.also {
+      // TODO create mappings here
```
andymarke, line 66: really doesn't do much other than gets each prisoners alerts and logs what we would do

mikehalmamoj (approved):

PR #561, opened 2024-04-12
SDIT-1614 Alert Migrate calls DPS and Mapping APIs

andymarke:
The DPS API accepts a list of alerts for the prisoner and that list is transformed into mapping that is POSTed in a single batch.

Since only one duplicate mapping can be duplicate the duplicate mapping handling differs for the usual pattern. This still requires a integration test

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsByPrisonerMigrationMappingApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsByPrisonerMigrationMappingApiService.kt
@@ -1,11 +1,35 @@
 package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.alerts
 
+import kotlinx.coroutines.reactive.awaitFirstOrDefault
 import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.core.ParameterizedTypeReference
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClient
+import org.springframework.web.reactive.function.client.WebClientResponseException
+import reactor.core.publisher.Mono
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.integration.history.CreateMappingResult
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.integration.history.DuplicateErrorResponse
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.integration.history.MigrationMapping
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsByPrisonerMigrationMappingApiService.kt
```diff
@@ -1,11 +1,35 @@
...
+import org.springframework.web.reactive.function.client.WebClientResponseException
+import reactor.core.publisher.Mono
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.integration.history.CreateMappingResult
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.integration.history.DuplicateErrorResponse
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.integration.history.MigrationMapping
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomismappings.model.AlertMappingDto
 
 @Service
 class AlertsByPrisonerMigrationMappingApiService(@Qualifier("mappingApiWebClient") webClient: WebClient) :
-  MigrationMapping<List<AlertMappingDto>>(domainUrl = "/mapping/alerts/all", webClient)
```
andymarke, line 11: had to override the default implementation that assumes the input to the service is the same as the duplicate output - in my case we send many mappings but the duplicate is only ever one of them

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/Transformer.kt
```diff
@@ -38,3 +39,21 @@ fun AlertResponse.toDPSMigratedAlert(offenderNo: String) = MigrateAlertRequest(
   updatedByDisplayName = this.audit.modifyDisplayName ?: this.audit.modifyUserId,
   comments = emptyList(),
 )
+
+fun AlertResponse.toDPSMigratedAlert(latestBooking: Boolean) = MigrateAlert(
+  offenderBookId = this.bookingId,
+  // TODO get real booking sequence
+  bookingSeq = if (latestBooking) 1 else 99,
```
mikehalmamoj, line 46: There's no unique constraint including bookingSeq is there?

andymarke: Nah, they won't be storing it - they really just need to know if it is 1 or not - anyway I will be fixing that in next PR before a do migration anyway

mikehalmamoj (approved):

PR #573, opened 2024-04-22
SDIT-1614 Use new prisoner centric mapping GET endpoint

andymarke:
This will give a count of prisoners migrated rather than alerts based on feature switch


lbennett-moj (dismissed):

petergphillips (approved):
LGTM

PR #581, opened 2024-04-26
SDIT-1696 Deletion locations sync events handled wrongly

steverendell:
deletion of a location attribute or usage should just result in an overall update


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/locations/LocationsSynchronisationService.kt
```diff
@@ -21,7 +21,7 @@ import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.InternalM
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.NomisApiService
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.SynchronisationQueueService
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.SynchronisationType.LOCATIONS
-import java.util.UUID
+import java.util.*
```
mikehalmamoj, line 24: Star imports! I think there's a bug in Intellij because I've noticed it happening to me too, and my settings are correct 🤷

mikehalmamoj (approved):

PR #587, opened 2024-04-30
SDIT-1568 Scaffolding to handle prisoner merges for alerts via domain…

andymarke:
… event

Also added call to nomis api to get alerts by booking id


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsEventListener.kt
```diff
@@ -17,7 +17,7 @@ import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.listeners.Synchro
 import java.util.concurrent.CompletableFuture
 
 @Service
-class AlertsPrisonOffenderEventListener(
```
andymarke, line 20: renamed to better reflect it also gets domain events

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/data/NomisPrisonerMergeEvent.kt
```diff
@@ -3,3 +3,14 @@ package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.data
...
 )
+
+data class PrisonerMergeDomainEvent(
+  val additionalInformation: MergeAdditionalInformationEvent,
+)
+
+data class MergeAdditionalInformationEvent(
+  val nomsNumber: String,
+  val removedNomsNumber: String,
+  val bookingId: Long?,
```
andymarke, line 14: these two booking ids do not exist on event right now, but I will change that NOW

petergphillips (approved):
LGTM

PR #612, opened 2024-05-14
SDIT-1730 Ignore alerts updated on previous bookings unless they are "relevant"

andymarke:
…

When mapping exists always update regardless but don't error if mapping exists, and it should not be in DPS anyway

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
@@ -89,13 +89,18 @@ class AlertsSynchronisationService(
     } else {
       val mapping = mappingApiService.getOrNullByNomisId(event.bookingId, event.alertSeq)
       if (mapping == null) {
-        telemetryClient.trackEvent(
-          "alert-synchronisation-updated-failed",
-          telemetry,
-        )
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
```diff
@@ -89,13 +89,18 @@ class AlertsSynchronisationService(
...
+          // if we do have a mapping update anyway even if it shouldn't have been in DPS - since this is edge case that is not
+          // significant enough to try to error on i.e. the scenario would have been it was migrated but since then a new alert
+          // has been created in DPS so we can just keep the old one forever rather than some weird logic of deleting it from DPS
+          telemetryClient.trackEvent("alert-synchronisation-updated-ignored-previous-booking", telemetry + ("bookingSequence" to nomisAlert.bookingSequence))
+        } else {
+          telemetryClient.trackEvent("alert-synchronisation-updated-failed", telemetry)
+          if (hasMigratedAllData) {
+            // after migration has run this should not happen so make sure this message goes in DLQ
+            throw IllegalStateException("Received ALERT-UPDATED for alert that has never been created")
+          }
```
mikehalmamoj, line 103: Couldn't see a test for this bit - sounds somewhat excepitonal though

mikehalmamoj (approved):

PR #617, opened 2024-05-24
SDIT-1753 Switch alert migration to use /prisoners/ids/all

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsNomisApiService.kt
```diff
@@ -38,22 +39,21 @@ class AlertsNomisApiService(@Qualifier("nomisApiWebClient") private val webClien
...
+      it.path("/prisoners/ids/all")
         .queryParam("page", pageNumber)
         .queryParam("size", pageSize)
-        .queryParam("active", false)
         .build()
     }
     .retrieve()
     .awaitBody()
 
-  suspend fun getAlertsToMigrate(offenderNo: String): PrisonerAlertsResponse = webClient.get()
```
andymarke, line 50: because we now get prisoners with no bookings we have to cope with 404 from this endpoint

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsByPrisonerMigrationService.kt
```diff
@@ -91,9 +91,7 @@ class AlertsByPrisonerMigrationService(
       telemetryClient.trackEvent(
         "alerts-migration-entity-migrated",
         mapOf(
-          "latestNomisBookingId" to latestNomisBookingId,
           "offenderNo" to offenderNo,
-          "status" to status,
           "migrationId" to context.migrationId,
```
andymarke, line 95: these fields have been dropped from endpoint for performance reasons  - they were used only for analytics

mikehalmamoj (approved):

PR #618, opened 2024-05-24
SDIT-1753 Reduce migration threads to 8

andymarke:
Reduce this problem of connection timeouts on 10 sized connection pool


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsByPrisonerMigrationMessageListener.kt
```diff
@@ -25,7 +25,12 @@ class AlertsByPrisonerMigrationMessageListener(
   alertsMigrationService,
 ) {
 
-  @SqsListener(ALERTS_QUEUE_ID, factory = "hmppsQueueContainerFactoryProxy")
+  @SqsListener(
+    ALERTS_QUEUE_ID,
+    factory = "hmppsQueueContainerFactoryProxy",
+    maxConcurrentMessages = "8",
```
andymarke, line 31: move from default of 10 down to 8

petergphillips (approved):
LGTM

PR #622, opened 2024-05-29
SDIT-1703 Split Circle tests into parallel executors

mikehalmamoj:
image

4 parallel executors roughly halved the time to test. Going up to 8 executors only shaved another 1-2 minutes off the total testing time. It seems that a lot of the time is taken up by Gradle rather than just tests running, hence the diminishing returns.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,6 +14,28 @@ parameters:
 jobs:
   validate:
     executor:
+      name: hmpps/java
+      tag: "21.0"
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - gradle-{{ checksum "build.gradle.kts" }}
+            - gradle-
+      - run:
+          command: ./gradlew check -x test
+      - save_cache:
+          paths:
+            - ~/.gradle
+          key: gradle-{{ checksum "build.gradle.kts" }}
+      - store_test_results:
+          path: build/test-results
+      - store_artifacts:
+          path: build/reports/tests
+
+  tests:
+    parallelism: 4
+    executor:
       name: hmpps/java_localstack_postgres
       jdk_tag: "21.0"
       localstack_tag: "3"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -13,6 +13,28 @@ parameters:
...
+      name: hmpps/java
+      tag: "21.0"
+    steps:
+      - checkout
+      - restore_cache:
+          keys:
+            - gradle-{{ checksum "build.gradle.kts" }}
+            - gradle-
+      - run:
+          command: ./gradlew check -x test
```
mikehalmamoj, line 26: Exclude tests from the validate job so we can run the tests in multiple parallel executors

.circleci/config.yml
```diff
@@ -28,7 +50,30 @@ jobs:
...
+          command: |
+            cd src/test/kotlin
+            
+            # Find all Kotlin files that contain tests, write the filenames to filenames.txt and tell CircleCI to split them by filesize across the parallel executors
+            circleci tests glob "**/*.kt" \
+            | xargs grep -l -E '@Test|@ParameterizedTest' \
+            | cut -c 1- \
+            | circleci tests run --command=">filenames.txt xargs echo" \
+            --verbose \
+            --split-by=filesize
```
mikehalmamoj, line 62: Would have preferred to split by timings for a more even split - unfortunately not possible because JUnit doesn't always report times against class names where @Nested or @DisplayName has been used.

mikehalmamoj: Will also investigate splitting the tests by package name and having a parallel executor for each. This would require handrolling an alternative to `circleci tests glob` but should give us a more intuitive solution and make it easier to spot naughty packages.

.circleci/config.yml
```diff
@@ -28,7 +50,30 @@ jobs:
...
+            
+            # Fail the build if no test files were found
+            [ -s filenames.txt ] || circleci-agent step halt
+      - run:
+          command: |
+            # Convert filenames for this parallel executor into packages/classes for Gradle test arguments
+            GRADLE_ARGS=$(cat src/test/kotlin/filenames.txt \
+            | sed 's@/@.@g' \
+            | sed 's/\.kt//g' \
+            | awk '{for (i=1; i<=NF; i++) print "--tests",$i}')
```
petergphillips, line 72:
```suggestion
            | awk '{print "--tests",$0}')
```

.circleci/config.yml
```diff
@@ -13,6 +13,28 @@ parameters:
...
+          keys:
+            - gradle-{{ checksum "build.gradle.kts" }}
+            - gradle-
+      - run:
+          command: ./gradlew check -x test
+      - save_cache:
+          paths:
+            - ~/.gradle
+          key: gradle-{{ checksum "build.gradle.kts" }}
+      - store_test_results:
```
petergphillips, line 31: don't think this step is required?

.circleci/config.yml
```diff
@@ -13,6 +13,28 @@ parameters:
...
+          path: build/test-results
+      - store_artifacts:
```
petergphillips, line 33: don't think there will be anything here either?

PR #625, opened 2024-05-31
SDIT-1703 Create a parallel executor per package

mikehalmamoj:
https://app.circleci.com/pipelines/github/ministryofjustice/hmpps-prisoner-from-nomis-migration/2400/workflows/779aca58-e36b-4ecd-9408-bc0670f31cbb/jobs/7323/parallel-runs/10?filterBy=ALL

image

image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,12 @@ parameters:
   releases-slack-channel:
     type: string
     default: dps-releases
+  test-executor-packages:
+    type: string
+    default: "activities adjudications alerts appointments courtsentencing csip incidents locations sentencing visits"
+  test-executors-count:
+    type: integer
+    default: 11
 
 jobs:
   validate:
@@ -30,7 +36,7 @@ jobs:
           key: gradle-{{ checksum "build.gradle.kts" }}
 
   tests:
-    parallelism: 4
+    parallelism: << pipeline.parameters.test-executors-count >>
     executor:
       name: hmpps/java_localstack_postgres
       jdk_tag: "21.0"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -10,6 +10,12 @@ parameters:
   releases-slack-channel:
     type: string
     default: dps-releases
+  test-executor-packages:
+    type: string
+    default: "activities adjudications alerts appointments courtsentencing csip incidents locations sentencing visits"
+  test-executors-count:
+    type: integer
+    default: 11
```
mikehalmamoj, line 18: List all packages in the test project that will run on a parallel executor. Any other packages found will be grouped toegether on another executor.

Note that the executor count must be 1 more than the list entries, and is declared here to hopefully make that obvious. There is validation to check the value is correct.

.circleci/config.yml
```diff
@@ -49,22 +55,75 @@ jobs:
...
+              # Create a file containing the classes for this dedicated executor package
+              echo "Finding test classes for dedicated executor from package=${dedicated_executor_packages[$CIRCLE_NODE_INDEX]}"
+              find_test_classes "${dedicated_executor_packages[$CIRCLE_NODE_INDEX]}" "$index_output_file"
+            else
+              # Create a single file to contain test classes bundled from all other packages
+              echo "Finding test classes for common executor from packages=${common_executor_packages[@]}"
+              for common_executor_package in "${common_executor_packages[@]}"; do
+                find_test_classes "$common_executor_package" "$index_output_file"
+              done
+            fi
```
mikehalmamoj, line 116: This is where the test splitting happens - each executor has a unique index `CIRCLE_NODE_INDEX` and for each one we curate a list of tests for the package with that index -> a unique output file.

The last executor gets the tests from any remaining packages.

.circleci/config.yml
```diff
@@ -49,22 +55,75 @@ jobs:
...
+            if [[ "$CIRCLE_NODE_INDEX" -lt "$((CIRCLE_NODE_TOTAL-1))" ]]; then
+              [ -s test_classes$CIRCLE_NODE_INDEX.txt ] || circleci-agent step halt
+            fi
       - run:
           command: |
-            # Convert filenames for this parallel executor into packages/classes for Gradle test arguments
-            GRADLE_ARGS=$(cat src/test/kotlin/filenames.txt \
-            | sed 's@/@.@g' \
-            | sed 's/\.kt//g' \
+            GRADLE_ARGS=$(cat src/test/kotlin/test_classes$CIRCLE_NODE_INDEX.txt \
```
mikehalmamoj, line 126: And here we just run the tests curated for this executor.

.circleci/config.yml
```diff
@@ -49,22 +55,75 @@ jobs:
...
             | awk '{for (i=1; i<=NF; i++) print "--tests",$i}')
```
mikehalmamoj, line 127: And here we just run the tests curated for this executor.

.circleci/config.yml
mikehalmamoj, line 126: And here we just run the tests curated for this executor.

.circleci/config.yml
```diff
@@ -49,22 +55,75 @@ jobs:
...
             
-            # Find all Kotlin files that contain tests, write the filenames to filenames.txt and tell CircleCI to split them by filesize across the parallel executors
-            circleci tests glob "**/*.kt" \
-            | xargs grep -l -E '@Test|@ParameterizedTest' \
-            | cut -c 1- \
-            | circleci tests run --command=">filenames.txt xargs echo" \
-            --verbose \
-            --split-by=filesize
+            # Find all packages            
+            all_packages=($(find uk/gov/justice/digital/hmpps/prisonerfromnomismigration -mindepth 1 -maxdepth 1 -type d -exec basename {} \;))
```
mikehalmamoj, line 59: This whole script isn't the easiest to follow. Not much ❤️ for bash in my mind. So I might move all the code into functions to make it easier to read. Next PR.

petergphillips: alternative to the `find` command would be something like
```
ls -p uk/gov/justice/digital/hmpps/prisonerfromnomismigration | grep '/' | tr -d '/'
```
since `ls -p` outputs a trailing `/` for directory items

andymarke (approved):
LGTM

.circleci/config.yml
mikehalmamoj: Thanks. If you have any other let me know and I'll implement when I move the code to functions.

PR #640, opened 2024-06-11
SDIT-1835 Ignore updates to adjustments deleted

andymarke:
It is possible to process an UPSERT event from NOMIS after an adjustment is deleted - this can happen if events are delayed. In this scenario just ignore the update rather than sending to DLQ


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/sentencing/SentencingAdjustmentsMigrationService.kt
```diff
@@ -77,8 +77,9 @@ class SentencingAdjustmentsMigrationService(
             nomisApiService.getKeyDateAdjustment(nomisAdjustmentId)
           }
 
+        // this service needs deleting now migration is done
```
andymarke, line 80: will do this in a separate PR

mikehalmamoj (approved):

PR #663, opened 2024-06-19
SDIT-1850 Recover gracefully if conflict received from DPS incidents api during migration

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/incidents/IncidentsMigrationService.kt
```diff
@@ -70,33 +71,68 @@ class IncidentsMigrationService(
...
+            mapOf(
+              "nomisIncidentId" to nomisIncidentId,
+              "dpsIncidentId" to it.id,
+              "migrationId" to migrationId,
+            ),
+          )
+        }
+      } catch (e: WebClientResponseException.Conflict) {
+        // We have a conflict - this should only ever happen if the incident was stored in DPS, but we didn't receive a response in time
+        // so is never added to the incidents mapping table
```
steverendell, line 94: Is there a code to check too? e.g. locations api sends errorCode == 107

steverendell (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/incidents/IncidentsMigrationService.kt
karenmillermoj: yeah I saw you had one of those - I have just checked - they have one (302) but I'm not sure if I need it - will take a look

PR #666, opened 2024-06-21
Implement physical attributes NOMIS to DPS synchronisation

mikehalmamoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/prisonperson/PrisonPersonSyncIntTest.kt
```diff
@@ -1,36 +1,439 @@
...
+                  ),
+                ),
+              ),
+            ),
+          ),
+        )
+
+        sendPhysicalAttributesChangedEvent()
+
+        verifyResults(type = "ignored", reason = "New physical attributes are empty")
```
andymarke, line 106: Question; are physical attributes not copied by NOMIS for new bookings (like it does for Alerts)?

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/prisonperson/PrisonPersonSyncIntTest.kt
```diff
@@ -1,36 +1,439 @@
...
+            assertThat(it["offenderNo"]).isEqualTo("A1234AA")
+            assertThat(it["bookingId"]).isEqualTo("12345")
+            assertThat(it["error"]).contains("Bad Request")
+          },
+          isNull(),
+        )
+      }
+
+      @Test
+      fun `should put message on DLQ if booking has no physical attributes`() = runTest {
```
andymarke, line 405: why is this an error?

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/prisonperson/PrisonPersonSyncIntTest.kt
mikehalmamoj: They are yes. This test is for a brand new booking - if you go into the physical info NOMIS screen it creates this "empty" record which makes no sense to synchronise.

andymarke (approved):
lgtm

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/prisonperson/PrisonPersonSyncIntTest.kt
mikehalmamoj: We shouldn't be receiving events for bookings without any physical attributes - but I chose not to assume that in the code, so had to handle the error scenario. Should never happen.

PR #671, opened 2024-06-24
SDIT-1859 Resynchronising alerts on receive to old bookings  

andymarke:
SDIT-1859 Resynchronising alerts on receive to old bookings  

When a prisoner is received on a old booking the alerts in DPS need resetting to match alerts in NOMIS therefore we need to 
* Get all current alerts as if this is a migration
* Replace alerts in DPS
* Replace mappings

TODO
* Call new DPS resynchronisation endpoint when available
* Narrow down scenario for old bookings that requires a search change


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsDpsApiMockServer.kt
```diff
@@ -99,11 +97,12 @@ class AlertsDpsApiMockServer : WireMockServer(WIREMOCK_PORT) {
     )
   }
 
-  fun stubMigrateAlert(
```
andymarke, line 102: some of these mocks no longer referenced so deleted (but makes diff look a bit odd)

lbennett-moj (approved):

PR #676, opened 2024-06-25
SDIT-1378 Page agencies to avoid throttling incidents api

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/incidents/IncidentsReconciliationService.kt
```diff
@@ -17,15 +18,19 @@ class IncidentsReconciliationService(
...
 ) {
   private companion object {
     val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
   suspend fun generateReconciliationReport(agencies: List<IncidentAgencyId>): List<MismatchIncidents> =
-    withContext(Dispatchers.Unconfined) {
-      agencies.map { async { checkIncidentsMatch(it.agencyId) } }
-    }.awaitAll().filterNotNull()
+    agencies.chunked(pageSize.toInt()).flatMap { pagedAgencies ->
```
karenmillermoj, line 29: The call to get all incident agency ids (which then use to call this method) could have been paged, but I didn't think it was needed so is now paged/chunked to avoid throttling the incidents api.  The list of agencies should be 120ish so this should still work ok.

andymarke (approved):
LGTM

PR #686, opened 2024-07-01
SDIT-1864 Switch to DPS resync endpoint for merges

andymarke:
Switch from using DPS merge endpoint to DPS resycn endpoint so that it is consistent with new booking etc.

This requires 2 calls to the DPS endpoint to replace all alerts on retained prisoner and remove all alerts fro removed prisoner.

The specialised version of PUT mappings merge prisoners replaces all mappings on retained prisoner  and removes mapping from removed prisoner (and this has a retry mechanism )


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
```diff
@@ -191,36 +192,40 @@ class AlertsSynchronisationService(
     val bookingId = prisonerMergeEvent.additionalInformation.bookingId
     val offenderNo = prisonerMergeEvent.additionalInformation.nomsNumber
     val removedOffenderNo = prisonerMergeEvent.additionalInformation.removedNomsNumber
-    val alerts = nomisApiService.getAlertsByBookingId(bookingId).alerts
-    val alertsMapping = alerts.groupBy { mappingApiService.getOrNullByNomisId(bookingId = it.bookingId, alertSequence = it.alertSequence) }
-    val newAlerts = alertsMapping[null] ?: emptyList()
-    val retainedDpsAlertIds = alertsMapping.keys.filterNotNull().map { it.dpsAlertId }
```
andymarke, line 197: meat of change, switch to different DPS endpoint and Mapping endpoint similar to how we handle new bookings etc

mikehalmamoj (approved):

PR #694, opened 2024-07-04
SDIT-1877 Process any CSIP deleted or CSIP Factor Delete - ignore auditModule

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/csip/CSIPPrisonOffenderEventListener.kt
```diff
@@ -71,6 +72,10 @@ class CSIPPrisonOffenderEventListener(
         RETRY_SYNCHRONISATION_MAPPING.name -> csipSynchronisationService.retryCreateCSIPReportMapping(
           sqsMessage.Message.fromJson(),
         )
+
+        RETRY_CSIP_FACTOR_SYNCHRONISATION_MAPPING -> csipFactorSynchronisationService.retryCreateCSIPFactorMapping(
```
karenmillermoj, line 76: Ensure pick up retries

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/csip/CSIPFactorSynchronisationService.kt
```diff
@@ -89,11 +90,6 @@ class CSIPFactorSynchronisationService(
         "nomisCSIPFactorId" to event.csipFactorId,
       )
 
-    if (event.auditModuleName == "DPS_SYNCHRONISATION") {
```
karenmillermoj, line 92: Previous PR comments highlighted we won't know where the delete was coming from - so this check is irrelevant

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/csip/CSIPPrisonOffenderEventListener.kt
andymarke: Are you missing a test here than? Where mapping create fails once? I think that is the only "unhappyish" path worth doing

andymarke (approved):
LGTM

PR #700, opened 2024-07-05
SDIT-1593: ✨ Switch to authentication holder from kotlin lib

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/AuditConfiguration.kt
```diff
@@ -1,15 +0,0 @@
...
-import org.springframework.data.domain.AuditorAware
-import org.springframework.stereotype.Service
-import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.SecurityUserContext
-import java.util.Optional
-
-@Configuration
-class AuditConfiguration
-
-@Service(value = "auditorAware")
-class AuditorAwareImpl(private val securityUserContext: SecurityUserContext) : AuditorAware<String> {
```
petergphillips, line 11: don't think this was actually used - is the non reactive version anyway

andymarke (approved):
lgtm

PR #711, opened 2024-07-10
SDIT-1886 Resynchronise alerts after a booking has moved.

andymarke:
Switched to using /mapping/alerts/{offenderNo}/all rather than the merge equivalent - the merge version would have simply removed mappings twice though did work

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
@@ -263,7 +263,7 @@ class AlertsSynchronisationService(
         },
       )
 
-      tryToReplaceMergedMappings(movedFromNomsNumber, MergedPrisonerAlertMappingsDto(movedFromNomsNumber, prisonerMappings), telemetry)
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
```diff
@@ -327,23 +327,6 @@ class AlertsSynchronisationService(
     }
   }
 
-  private suspend fun tryToCreateMappings(
```
andymarke, line 330: unrelated change - wasn't being used

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
```diff
@@ -263,7 +263,7 @@ class AlertsSynchronisationService(
         },
       )
 
-      tryToReplaceMergedMappings(movedFromNomsNumber, MergedPrisonerAlertMappingsDto(movedFromNomsNumber, prisonerMappings), telemetry)
+      tryToReplaceMappings(offenderNo = movedFromNomsNumber, prisonerMappings = prisonerMappings, telemetry)
       telemetryClient.trackEvent(
```
andymarke, line 267: slightly different version of mapping endpoint

lbennett-moj (approved):

PR #721, opened 2024-07-18
SDIT-1901 From Nomis Locations attribute or usage events should only update

steverendell:
atrribute or usage changes in Nomis should not cause an attempt to create the location in DPS


Mjwillis (approved):
LGTM

PR #727, opened 2024-07-22
SDIT-1825 Allow testing with single prisoner

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/prisonperson/PrisonPersonMigrationService.kt
```diff
@@ -54,10 +55,16 @@ class PrisonPersonMigrationService(
...
+  ): PageImpl<PrisonerId> =
+    if (migrationFilter.prisonerNumber == null) {
+      nomisService.getPrisonerIds(
+        pageNumber = pageNumber,
+        pageSize = pageSize,
+      )
+    } else {
+      // If a single prisoner migration is requested then we must be testing. Pretend that we called nomis-prisoner-api which found a single prisoner.
+      PageImpl<PrisonerId>(mutableListOf(PrisonerId(migrationFilter.prisonerNumber)), Pageable.ofSize(1), 1)
+    }
```
mikehalmamoj, line 67: Couldn't be bothered to write a new NOMIS API endpoint to add a prisoner number filter to getIds just so I can test migration of single prisoners.

If we had a new getIds endpoint with a prisoner number filter and the prisoner didn't exist then it would find nothing and there'd be an empty migration.

By stubbing the getIds call instead, if the prisoner doesn't exist then we'll get a 404 from NOMIS API when we try to get the data, and the message will end up on the DLQ. Which is fine.

andymarke (approved):
LGTM

PR #745, opened 2024-08-01
SDIT-1924 Add _JAVA_OPTIONS to circle build and spring boot upgrade

karenmillermoj:


.circleci/config.yml
```diff
@@ -30,6 +38,13 @@ jobs:
             - gradle-
       - run:
           command: ./gradlew check -x test
+      - persist_to_workspace:
+          root: .
+          paths:
+            - build
+            - src
+            - build.gradle.kts
```
karenmillermoj, line 46: I wasn't sure which ones to add here - src and build.gradle.kts were added when testing - so I will remove

.circleci/config.yml
```diff
@@ -45,9 +60,19 @@ jobs:
       postgres_tag: "16"
       postgres_username: "migration"
       postgres_password: "migration"
+    environment:
+      # Setting max gradle workers to 1 (which means 1 running concurrently, not max 1!) - because although this doesn't stop the ktlint workers from starting it does seem to prevent the test executor from needing as much memory.
+      _JAVA_OPTIONS: -Xmx1g -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2
```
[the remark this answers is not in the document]
petergphillips: also worth removing the `resource_class` line as not needed since split into 12 parallel test runs

.circleci/config.yml
```diff
@@ -37,14 +37,21 @@ jobs:
 
   tests:
     parallelism: << pipeline.parameters.test-executors-count >>
-    resource_class: large
```
petergphillips, line 40: can't see this is needed, think it was required when there was just one job

mikehalmamoj (approved):

PR #752, opened 2024-08-07
SDIT-1929 Booking move now updates both parties

andymarke:
When the "from" prisoner is inactive resynchronise alerts. When they are active the alerts would have resynchronised anyway due to the new active booking.


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationIntTest.kt
```diff
@@ -1170,7 +1171,7 @@ class AlertsSynchronisationIntTest : SqsIntegrationTestBase() {
             movedToNomsNumber = movedToNomsNumber,
           ),
         )
-        waitForAnyProcessingToComplete("from-nomis-synch-alerts-booking-moved")
+        waitForAnyProcessingToComplete("from-nomis-synch-alerts-booking-moved", "from-nomis-synch-alerts-booking-moved-ignored")
       }
```
andymarke, line 1175: by default it won't process both prisoners, so wait for both events

mikehalmamoj (approved):

PR #755, opened 2024-08-08
SDIT-1930: ✨ Latest sqs and add in eventType for sent messages

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/SynchronisationQueueService.kt
```diff
@@ -33,6 +34,9 @@ class SynchronisationQueueService(
       SendMessageRequest.builder()
         .queueUrl(queue.queueUrl)
         .messageBody(sqsMessage.toJson())
+        .messageAttributes(
+          mapOf("eventType" to MessageAttributeValue.builder().dataType("String").stringValue("prisoner-from-nomis-synchronisation-$messageType").build()),
```
petergphillips, line 38: I'm confused as to the difference between this message and the one from the migration queue service - are they both synchronisation messages?

andymarke: They will go to different queues though the scenario will be similar.
1) An error create a mapping while running a bulk migration
2) An error creating a mapping (and other retries) while processing a NOMIS xtag event

andymarke (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/SynchronisationQueueService.kt
petergphillips: turns out the migration one was never used, so have removed it

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/MigrationQueueService.kt
```diff
@@ -52,23 +55,6 @@ class MigrationQueueService(
     }
   }
 
-  suspend fun <T : Enum<T>> sendMessage(message: T, context: SynchronisationContext<*>) {
```
petergphillips, line 55: this did feel weird - a migration service using a synchronisation context, but turns out it wasn't used

PR #756, opened 2024-08-09
SDIT-1931 Set token retrieval timeout

andymarke:
Also added test to expose issue and demonstrate fix. Test can be deleted once fix is incorporated into kotlin template library

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/config/WebClientConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/config/WebClientConfiguration.kt
@@ -3,8 +3,15 @@ package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.config
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
+import org.springframework.http.client.reactive.ReactorClientHttpConnector
+import org.springframework.security.oauth2.client.AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager
 import org.springframework.security.oauth2.client.ReactiveOAuth2AuthorizedClientManager
+import org.springframework.security.oauth2.client.ReactiveOAuth2AuthorizedClientProviderBuilder
+import org.springframework.security.oauth2.client.ReactiveOAuth2AuthorizedClientService
+import org.springframework.security.oauth2.client.endpoint.WebClientReactiveClientCredentialsTokenResponseClient
+import org.springframework.security.oauth2.client.registration.ReactiveClientRegistrationRepository
 import org.springframework.web.reactive.function.client.WebClient
+import reactor.netty.http.client.HttpClient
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/config/WebClientConfiguration.kt
```diff
@@ -34,4 +41,31 @@ class WebClientConfiguration(
...
+
+  @Bean
+  fun reactiveOAuth2AuthorizedClientManagerWithTimeout(
+    reactiveClientRegistrationRepository: ReactiveClientRegistrationRepository,
+    reactiveOAuth2AuthorizedClientService: ReactiveOAuth2AuthorizedClientService,
+  ): ReactiveOAuth2AuthorizedClientManager = AuthorizedClientServiceReactiveOAuth2AuthorizedClientManager(
+    reactiveClientRegistrationRepository,
+    reactiveOAuth2AuthorizedClientService,
+  ).apply {
+    val accessTokenResponseClient =
```
andymarke, line 53: this really the only difference to the bean in the library, manually add a client with a timeout, the standard code just calls WebClient.builder().build()

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationPerformanceTest.kt
```diff
@@ -0,0 +1,94 @@
+package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.alerts
```
andymarke, line 1: Over the top test that we can delete in the future, I just wanted to reproduce the exact scenario before applying fix

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/config/WebClientConfiguration.kt
petergphillips: looks a lot of code to have to write to override, I'll have a play and see if there is a simpler way

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/config/WebClientConfiguration.kt
petergphillips: the simpler way is to add the code to the library instead 😂

andymarke: 100% - but want to test it first.

BTW I'm not sure if it is as easy as you would imagine @petergphillips 

The bean creation here has access to the required timeout value, the bean creating in the library is separated from the main webClient creation that knows about the timeout.

I think I will leave it for @mikehalmamoj  - since he will probably want some fancy way of coinfiguring the auth timeout for each webClient where as my solution sets each web client auth client to have the same timeout since that is exactly what we want. Others my have other requirements.

PR #771, opened 2024-08-22
SDIT-1985 Set 5 seconds between Incident events

karenmillermoj:


src/main/resources/application-localstack.yml
```diff
@@ -54,10 +54,10 @@ hmpps.sqs:
     eventincidents:
       queueName: ${random.uuid}
       dlqName: ${random.uuid}
+      errorVisibilityTimeout: 0
```
karenmillermoj, line 57: Too much context switching - added against csip instead of incidents (now corrected)

petergphillips (approved):
LGTM

PR #775, opened 2024-08-22
SDIT-194 Handle CSIP_ATTENDEES-DELETED event

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/csip/CSIPDpsApiService.kt
```diff
@@ -72,6 +72,14 @@ class CSIPDpsApiService(@Qualifier("csipApiWebClient") private val webClient: We
       .awaitBodilessEntity()
   }
 
+  suspend fun deleteCSIPAttendee(csipAttendeeId: String) {
```
karenmillermoj, line 75: After further investigation - this change may not be needed - however I have completed the PR incase we do need it and also there is some name refactoring

petergphillips (approved):
LGTM

PR #797, opened 2024-09-09
SDIT-2033 Delete csip child table event capture is not required

karenmillermoj:


helm_deploy/values-preprod.yaml
```diff
@@ -38,19 +38,14 @@ generic-service:
     FEATURE_EVENT_CSIP_REPORTS-DELETED: false
```
karenmillermoj, line 38: Code relating to the actual child table deletes will be removed in a future PR, once all delete code (for top level CSIP) is complete

mikehalmamoj (approved):

PR #840, opened 2024-09-23
SDIT-1898 Switch mock casenotes-api to real one

steverendell:
Add CNOTE/OUTCOME correction. A few old case notes in Nomis have an invalid type/subtype combination.


petergphillips (approved):
LGTM

PR #842, opened 2024-09-24
SDIT-2019 Send all profile types to DPS on each request

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/prisonperson/profiledetails/ProfileDetailsSyncService.kt
```diff
@@ -103,8 +103,8 @@ class ProfileDetailsSyncService(
       null
     }
 
-  private fun ProfileDetailsResponse.toDpsRequestIfType(profileType: String) =
-    takeIf { type == profileType }
+  private fun List<ProfileDetailsResponse>.toDpsRequest(profileType: String) =
+    find { it.type == profileType }
```
mikehalmamoj, line 107: This is the main change - send all profile details and not just the one that has changed. This is to work around problems in the service we're callign where requests are overwriting each other - so now we just send all data in all requests.

andymarke (approved):
lgtm

PR #846, opened 2024-09-25
SDIT-1940 Add capture of insert/update events for attendees/interviews/plans/reviews

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/csip/SyncTransformer.kt
```diff
@@ -114,7 +114,14 @@ fun CSIPResponse.toActionDetails() = getActionDetails(
   lastModifiedBy,
   lastModifiedByDisplayName,
 )
-
+fun Attendee.toActionDetails() = getActionDetails(
```
karenmillermoj, line 117: There is no inheritance here so couldn't figure out a neat solution apart from all calling the same getter

andymarke: I suppose you could add an interface to all JPA objects with the audit columns  but not sure I would bother

andymarke (approved):
LGTM

PR #858, opened 2024-10-04
SDIT-2060: 🐛 Propagate case notes tracing and upgrade sqs to fix missing RECEIVE

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/SynchronisationQueueService.kt
```diff
@@ -20,7 +21,7 @@ class SynchronisationQueueService(
   private val objectMapper: ObjectMapper,
 ) {
 
-  fun sendMessage(messageType: String, synchronisationType: SynchronisationType, message: Any, telemetryAttributes: Map<String, String> = mapOf(), delaySeconds: Int = 0) {
+  suspend fun sendMessage(messageType: String, synchronisationType: SynchronisationType, message: Any, telemetryAttributes: Map<String, String> = mapOf()) {
```
petergphillips, line 24: this was missing the `suspend` and `await`, `delaySeconds` unused

src/main/resources/application.yml
```diff
@@ -175,8 +175,6 @@ azure:
 
 hmpps.sqs:
   queues:
-    migrationcasenotes:
-      propagateTracing: false
```
petergphillips, line 179: turn on tracing for case notes now that we are not tracing through individual entities

steverendell (approved):
lgtm

PR #861, opened 2024-10-07
Revert "SDIT-2019 Send all profile types to DPS on each request (#842)"

mikehalmamoj:
This reverts commit 0ad8b07e18a99753a7bbf20287fbc8915a2599e8.


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/prisonperson/profiledetails/ProfileDetailsSyncService.kt
```diff
@@ -103,8 +103,8 @@ class ProfileDetailsSyncService(
       null
     }
 
-  private fun List<ProfileDetailsResponse>.toDpsRequest(profileType: String) =
-    find { it.type == profileType }
+  private fun ProfileDetailsResponse.toDpsRequestIfType(profileType: String) =
+    takeIf { type == profileType }
```
mikehalmamoj, line 107: We put a temporary workaround in for DPS concurrent updates problem to send all details at the same time. Now that concurrent updates are fixed we can go back to sending each profile type individually.

andymarke (approved):
lgtm

PR #909, opened 2024-10-30
SDIT-2090 Pre-work for calling DPS migrate endpoint

andymarke:
Use very old version of DPS migrate endpoint and change tests to use it.

TODO - new version of endpoint to be released very soon


petergphillips (approved):
LGTM

PR #910, opened 2024-10-30
SDIT-2172: 🐛 Ensure that futures are awaited for and avoid thenAccept

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/listeners/ListenerHelpers.kt
```diff
@@ -23,5 +23,8 @@ data class SQSMessage(val Type: String, val Message: String, val MessageId: Stri
 data class MessageAttributes(val eventType: EventType)
 data class EventType(val Value: String, val Type: String)
 
-fun asCompletableFuture(process: suspend () -> Unit): CompletableFuture<Void> =
-  CoroutineScope(Context.current().asContextElement()).future { process() }.thenAccept { }
```
petergphillips, line 27: I think the `thenAccept{}` here was used to ensure we got a `Void` back, but can do that without chaining the future

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/MigrationQueueService.kt
```diff
@@ -52,7 +52,7 @@ class MigrationQueueService(
         .eventTypeMessageAttributes("prisoner-from-nomis-migration-${context.type.telemetryName}", noTracing = noTracing)
         .delaySeconds(delaySeconds)
         .build(),
-    ).thenAccept {
+    ).await().also {
```
petergphillips, line 55: two separate issues here
1. We weren't waiting for the message to be sent, which is why I think the spans / threads were then getting mixed up
2. The `trackEvent` message was happening in a different future which meant that the `OperationId` etc. was missing

mikehalmamoj (approved):

PR #916, opened 2024-11-01
SDIT-2173 Test to check only mandatory fields are required for contacts

andymarke:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/contactperson/ContactPersonDpsApiMockServer.kt
```diff
@@ -28,7 +29,10 @@ class ContactPersonDpsApiExtension :
     @JvmField
     val dpsContactPersonServer = ContactPersonDpsApiMockServer()
     lateinit var objectMapper: ObjectMapper
+
+    @Suppress("unused")
     inline fun <reified T> getRequestBody(pattern: RequestPatternBuilder): T = dpsContactPersonServer.getRequestBody(pattern, objectMapper)
```
andymarke, line 34: I'll keep this since it asserts that is at least one request find - so is of use as a convenience when you know there should be just one request made to a stub

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/contactperson/ContactPersonMigrationIntTest.kt
```diff
@@ -283,6 +303,26 @@ class ContactPersonMigrationIntTest : SqsIntegrationTestBase() {
...
+        assertThat(request.title).isNull()
+        assertThat(request.language).isNull()
+        assertThat(request.interpreterRequired).isFalse()
+        assertThat(request.domesticStatus).isNull()
+        assertThat(request.deceasedDate).isNull()
+        assertThat(request.staff).isFalse()
+        assertThat(request.createUsername).isEqualTo("ADJUA.BEEK")
+        assertThat(request.createDateTime).isEqualTo(LocalDateTime.parse("2022-01-02T10:23"))
+        assertThat(request.modifyUsername).isNull()
+        assertThat(request.modifyDateTime).isNull()
```
mikehalmamoj, line 324: You could lose some nosie with:
```
with(requests.find { it.personId == 2000L }!!) {
  assertThat(firstName).isEqualTo("KWAME")
  ...etc
}
```
You lose the reference to `request` everywhere and the !! would fail the test if request not found.

mikehalmamoj (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/contactperson/ContactPersonMigrationIntTest.kt
mikehalmamoj: Or you could continue to throw for the nice error message 🤷

mikehalmamoj (approved):

PR #933, opened 2024-11-08
SDIT-2199 Allow re-sync of existing incident

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/incidents/IncidentsDataRepairResource.kt
```diff
@@ -0,0 +1,37 @@
...
+) {
+  @PostMapping("/incidents/{incidentId}/repair")
+  @ResponseStatus(HttpStatus.OK)
+  @PreAuthorize("hasRole('ROLE_MIGRATE_INCIDENTS')")
+  @Operation(
+    summary = "Resynchronises an existing incident from NOMIS back to DPS",
+    description = "Used when an unexpected event has happened in NOMIS that has resulted in the DPS data drifting from NOMIS, so emergency use only. Requires ROLE_MIGRATE_INCIDENTS",
+  )
+  suspend fun repairIncident(@PathVariable incidentId: Long) {
+    incidentsSynchronisationService.synchroniseIncidentUpdate(
```
karenmillermoj, line 25: Have left as an incident update only (wasn't sure if we needed create, we don't, for now)

PR #957, opened 2024-11-21
SDIT-2200 Implement move booking happy path for height and weight

mikehalmamoj:


helm_deploy/values-dev.yaml
```diff
@@ -25,6 +25,7 @@ generic-service:
     API_BASE_URL_PRISON_PERSON: https://prison-person-api-dev.prison.service.justice.gov.uk
     API_BASE_URL_CONTACT_PERSON: https://contacts-api-dev.hmpps.service.justice.gov.uk
     FEATURE_EVENT_COURT_EVENT_CHARGES-INSERTED: false
+    FEATURE_EVENT_PRISONPERSON_PRISON-OFFENDER-EVENTS_PRISONER_BOOKING_MOVED: false
```
mikehalmamoj, line 28: I think this means the event is still consumed by alerts but not by prison person.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/prisonperson/PrisonPersonMoveBookingService.kt
```diff
@@ -0,0 +1,80 @@
...
+      "toOffenderNo" to toOffenderNo,
+      "fromOffenderNo" to fromOffenderNo,
+    )
+
+    try {
+      syncFromOffenderDps(fromOffenderNo, telemetry)
+
+      syncToOffenderDps(toOffenderNo, telemetry)
+
+      syncToOffenderNomis(toOffenderNo, telemetry)
```
mikehalmamoj, line 35: This is where the action happens...

- sync the "from offender" to DPS - DPS will have been updated with wrong values from the to offender's booking
- sync "to offender" to DPS where a value has changed - DPS retains the values from the previous booking, so needs a sync for any fields that were updated
- sync "to offender" back to NOMIS - the NOMIS booking that was moved had incorrect values copied from the "from offender" so udpate them

helm_deploy/values-preprod.yaml
```diff
@@ -38,6 +38,7 @@ generic-service:
     FEATURE_EVENT_OFFENDER_SENTENCES-INSERTED: false
     FEATURE_EVENT_OFFENDER_SENTENCES-DELETED: false
     FEATURE_EVENT_OFFENDER_SENTENCES-UPDATED: false
+    FEATURE_EVENT_PRISONPERSON_PRISON-OFFENDER-EVENTS_PRISONER_BOOKING_MOVED: false
```
mikehalmamoj, line 41: Turn off only for prison person in preprod (and prod below). Turned on in dev so I can test height/weight works before moving onto the profile details.

andymarke (approved):
Lgtm

PR #962, opened 2024-11-22
SDIT-2248 Revised case notes merge strategy

steverendell:
New logic in the case notes merge which avoids migrating duplicate case notes created by merges, but including mappings for them pointing to the same DPS id as the original case notes


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/casenotes/CaseNotesMappingApiService.kt
```diff
@@ -1,38 +1,17 @@
...
-    webClient.post()
-      .uri("/mapping/casenotes/batch")
-      .bodyValue(mappings)
-      .retrieve()
-      .bodyToMono(Unit::class.java)
-      .map { CreateMappingResult<CaseNoteMappingDto>() }
-      .onErrorResume(WebClientResponseException.Conflict::class.java) {
-        Mono.just(CreateMappingResult(it.getResponseBodyAs(errorJavaClass)))
-      }
-      .awaitFirstOrDefault(CreateMappingResult())
```
steverendell, line 34: Unused

mikehalmamoj (approved):

PR #971, opened 2024-11-28
SDIT-2200 Handle scenario where ALL details updated in DPS before booking moved

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/prisonperson/PrisonPersonMoveBookingService.kt
```diff
@@ -56,7 +56,12 @@ class PrisonPersonMoveBookingService(
       nomisResponse.bookings
         .find { it.latestBooking }
         ?.also { latestBooking ->
-          physicalAttributesSyncService.physicalAttributesChanged(fromOffenderNo, latestBooking.bookingId, nomisResponse)
+          physicalAttributesSyncService.physicalAttributesChanged(
+            offenderNo = fromOffenderNo,
+            bookingId = latestBooking.bookingId,
+            nomisPhysicalAttributes = nomisResponse,
+            forceSync = true,
+          )
```
mikehalmamoj, line 64: I only did half a fix last time - if the offender had been updated in NOMIS. Didn't take into account that all of the details could have been entered in NOMIS, in which case the sync is ignored - so we always force the sync from NOMIS to DPS for the from offender too.

andymarke (approved):
lgtm

PR #989, opened 2024-12-06
SDIT-2341 Revised case notes from-nomis delete sync logic

steverendell:


petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/casenotes/CaseNotesSynchronisationService.kt
```diff
@@ -155,6 +155,21 @@ class CaseNotesSynchronisationService(
     try {
       caseNotesMappingService.getMappingGivenNomisIdOrNull(event.caseNoteId)
         ?.also { mapping ->
+          caseNotesMappingService.getByDpsId(mapping.dpsCaseNoteId)
+            .filter { it.nomisCaseNoteId != event.caseNoteId }
```
petergphillips, line 159: why filter out these records?

PR #999, opened 2024-12-10
SDIT-2345 flaky test in from-nomis

steverendell:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/casenotes/CaseNotesMergeIntTest.kt
```diff
@@ -139,6 +139,18 @@ class CaseNotesMergeIntTest : SqsIntegrationTestBase() {
...
+        verify(telemetryClient).trackEvent(
+          eq("casenotes-prisoner-merge"),
+          check {
+            assertThat(it["offenderNo"]).isEqualTo("A1234BB")
+            assertThat(it["removedOffenderNo"]).isEqualTo("A1234AA")
+            assertThat(it["bookingId"]).isEqualTo("1")
+          },
+          isNull(),
+        )
+      }
```
steverendell, line 153: Problem was with the next test starting before the service call from the previous one had finished. So the previous call clocked up stub hits making the verify call count wrong.

petergphillips (approved):
LGTM

PR #1020, opened 2024-12-23
SDIT-2290 offenderIdDisplay can be null

andymarke:
It can be null during a cascading delete where the contact has already been deleted


petergphillips (approved):
LGTM

PR #1024, opened 2024-12-31
SDIT-2366-Case-notes-migration-confused-by-identical-case-notes

steverendell:
Revised migration algorithm using groupBy


andymarke (approved):
lgtm

PR #1025, opened 2024-12-31
SDIT-2043 Used "Entered By" username for restrictions

andymarke:
DPS wants the enteredBy username for either the modified or created username since this is more likely to correspond to the real user that created or modified the restriction (for instance it survives a merge)

This would only affect a tiny percentage of contact restrictions but is needed since DPS has dropped the notion of the separate "enteredBy" field and rely on the audit fields


petergphillips (approved):
LGTM

PR #1055, opened 2025-01-20
SDIT-2395 Pick up api change for offender identifiers

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/coreperson/Transformer.kt
```diff
@@ -46,3 +71,13 @@ fun OffenderPhoneNumber.toCprPhoneNumber() =
     phoneType = type.code,
     phoneExtension = extension,
   )
+fun OffenderBelief.toCprBelief() = MockCprBelief(
+  nomisBeliefId = beliefId,
+  religion = belief.description,
```
petergphillips, line 76: I think some of these should be `code` instead?

karenmillermoj: yes - I agree - I saw the comment yesterday that the cpr team made and mentally made a note - but then physically forgot

PR #1060, opened 2025-01-22
SDIT-2480 Remove all exclude program codes logic

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/activities/ActivitiesMigrationService.kt
```diff
@@ -53,7 +53,6 @@ class ActivitiesMigrationService(
   ): PageImpl<FindActiveActivityIdsResponse> =
     nomisApiService.getActivityIds(
       prisonId = migrationFilter.prisonId,
-      excludeProgramCodes = activitiesApiService.getActivityCategories(),
```
mikehalmamoj, line 56: The main change is that we no longer use the excluded program codes in nomis prisoner api, and so we no longer need the activity categories.

lbennett-moj (approved):
lgtm

PR #1081, opened 2025-02-07
sdit-2538 adjust case reference delete behavour

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingSynchronisationService.kt
```diff
@@ -542,19 +542,16 @@ class CourtSentencingSynchronisationService(
             telemetry,
           )
         } ?: let {
-          // TODO determine whether retry is the best option here
           telemetryClient.trackEvent(
-            "court-charge-synchronisation-deleted-failed",
-            telemetry,
+            "court-charge-synchronisation-deleted-skipped",
+            telemetry + ("reason" to "charge is not mapped"),
```
lbennett-moj, line 547: we do get delete events after parent objects have gone.   This is unlikely to happen before creation of the parent event unless there was a big delay, plus the reconciliation would pick it up

PR #1098, opened 2025-02-12
Sdit 2448 corporate deleted

andymarke:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/integration/SqsIntegrationTestBase.kt
```diff
@@ -257,7 +255,7 @@ class SqsIntegrationTestBase : TestBase() {
 }
 
 internal fun SqsAsyncClient.sendMessage(queueOffenderEventsUrl: String, message: String) = sendMessage(SendMessageRequest.builder().queueUrl(queueOffenderEventsUrl).messageBody(message).build()).get()
-
+internal fun HmppsQueue.sendMessage(message: String) = this.sqsClient.sendMessage(this.queueUrl, message = message)
```
andymarke, line 258: avoid need to exposing url and client - only used in new test so far

PR #1101, opened 2025-02-13
SDIT-2442 Handle ADDRESSES_CORPORATE-INSERTED

andymarke:
and refactor Mapping creation to reduce shared code

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/organisations/OrganisationsMappingApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/organisations/OrganisationsMappingApiService.kt
@@ -30,7 +30,7 @@ class OrganisationsMappingApiService(@Qualifier("mappingApiWebClient") webClient
     }
     .awaitFirstOrDefault(CreateMappingResult())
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/organisations/OrganisationsMappingApiService.kt
```diff
@@ -30,7 +30,7 @@ class OrganisationsMappingApiService(@Qualifier("mappingApiWebClient") webClient
     }
     .awaitFirstOrDefault(CreateMappingResult())
 
-  suspend fun createOrganisationMapping(mapping: CorporateMappingDto): CreateMappingResult<CorporateMappingDto> = createMapping("/mapping/corporate/organisation", mapping)
+  suspend fun createOrganisationMapping(mapping: CorporateMappingDto): CreateMappingResult<OrganisationMapping> = createMapping("/mapping/corporate/organisation", mapping)
```
andymarke, line 33: All mappings related to organisations always return an mapping that has a dpsId and nomisId - so this allows us to share the duplicate handling code across different DTOs that have no common class but all have the same "shape"

mikehalmamoj (approved):

PR #1114, opened 2025-02-18
SDIT-2435 Handle INTERNET_ADDRESSES_CORPORATE-UPDATED

andymarke:
Double complication:
NOMIS has a single event and table for email and web but DPS and Mapping have separate endpoints; so a little bit more complex when mapping

Also in NOMIS a web address can be converted to email address and vice versa; in these scenarios we need to perform a DELETE in DPS and mapping service followed by creates

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/organisations/OrganisationsSynchronisationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/organisations/OrganisationsSynchronisationService.kt
@@ -389,6 +389,77 @@ class OrganisationsSynchronisationService(
     }
   }
   suspend fun corporateInternetAddressUpdated(event: CorporateInternetAddressEvent) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/organisations/OrganisationsSynchronisationService.kt
```diff
@@ -573,7 +644,19 @@ fun CorporateInternetAddress.toDpsCreateOrganisationEmailRequest(dpsOrganisation
   createdBy = this.audit.createUsername,
   createdTime = this.audit.createDatetime.toDateTime(),
 )
-fun CorporateInternetAddress.toDpsUpdateOrganisationEmailRequest(dpsOrganisationId: Long) = SyncUpdateOrganisationEmailRequest(
+fun CorporateInternetAddress.toDpsCreateOrganisationEmailRequestForUpdateSwitch(dpsOrganisationId: Long) = SyncCreateOrganisationEmailRequest(
+  emailAddress = this.internetAddress,
+  organisationId = dpsOrganisationId,
+  createdBy = this.audit.modifyUserId!!,
```
andymarke, line 650: this is a DPS create from a NOMIS update

karenmillermoj (approved):
lgtm

PR #1118, opened 2025-02-19
SDIT-2442 INTERNET_ADDRESSES_CORPORATE-UPDATED no mapping scenario

andymarke:
Since there can either be a web or email mapping we need to specifically deal when neither mapping is found and throw exception

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/organisations/OrganisationsSynchronisationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/organisations/OrganisationsSynchronisationService.kt
@@ -429,36 +429,43 @@ class OrganisationsSynchronisationService(
             )
           }
         }
-      }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/organisations/OrganisationsSynchronisationService.kt
```diff
@@ -429,36 +429,43 @@ class OrganisationsSynchronisationService(
...
-            )
-          } else {
-            dpsApiService.updateOrganisationEmail(
-              mapping.dpsId.toLong(),
-              nomisInternetAddress.toDpsUpdateOrganisationEmailRequest(),
-            )
           }
         }
+      } ?: run {
+        telemetryClient.trackEvent(
```
andymarke, line 464: previously this would silently fail and consume message

lbennett-moj (dismissed):

mikehalmamoj (approved):

PR #1134, opened 2025-02-25
sdit-1807 integrate sentence create

lbennett-moj:


helm_deploy/values-dev.yaml
```diff
@@ -28,6 +28,9 @@ generic-service:
     API_BASE_URL_PERSONAL_RELATIONSHIPS: https://personal-relationships-api-dev.hmpps.service.justice.gov.uk
     API_BASE_URL_ORGANISATIONS: https://organisations-api-dev.hmpps.service.justice.gov.uk
     FEATURE_EVENT_PERSONALRELATIONSHIPS_OFFENDER_PHYSICAL_DETAILS-CHANGED: false
+    FEATURE_EVENT_OFFENDER_SENTENCES-INSERTED: false
+    FEATURE_EVENT_OFFENDER_SENTENCES-DELETED: false
+    FEATURE_EVENT_OFFENDER_SENTENCES-UPDATED: false
```
lbennett-moj, line 34: these don't look right - think I'm missing the queue/organisation. Will revisit in next commit, the events aren't being emitted yet

andymarke: They are fine; you only need additional "domain" key if there is likely to be a conflict with another project; e.g somebody else was interested in OFFENDER_SENTENCES-*

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingSynchronisationService.kt
```diff
@@ -654,7 +654,7 @@ class CourtSentencingSynchronisationService(
     nomisSentence.offenderCharges.map {
       mappingApiService.getOffenderChargeByNomisId(it.id).dpsCourtChargeId
     }
-  } catch (notFoundException: WebClientResponseException.NotFound) {
+  } catch (webResponseException: WebClientResponseException) {
```
andymarke, line 657: Shouldn't this only be NotFound? a 500 exception I guess shouldn't result in ParentEntityNotFoundRetry you can just allow normal error processing else it might be confusing

andymarke (approved):
LGTM

PR #1139, opened 2025-02-27
SDIT-2629 The sync endpoint always syncs the latest booking

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/personalrelationships/profiledetails/ContactPersonProfileDetailsResource.kt
```diff
@@ -23,10 +23,10 @@ import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.config.ErrorRespo
 class ContactPersonProfileDetailsResource(
   private val syncService: ContactPersonProfileDetailsSyncService,
 ) {
-  @PutMapping("/{prisonerNumber}/{bookingId}/{profileType}")
+  @PutMapping("/{prisonerNumber}/{profileType}")
```
mikehalmamoj, line 26: After testing I realised that the sync endpoint should always sync the latest booking, so no need to pass in the booking ID in that case. (as apposed to processing a sync event for which the booking ID is provided - currently ignored if not latest booking but possibly switching to include old bookings in the future).

Everything in this PR is to cater for a possible missing booking ID.

petergphillips (approved):
LGTM

PR #1143, opened 2025-02-28
SDIT-2517 Send DPS location ID instead of from NOMIS

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/activities/ActivitiesMigrationService.kt
```diff
@@ -149,24 +150,24 @@ class ActivitiesMigrationService(
...
+  private suspend fun GetActivityResponse.toActivityMigrateRequest(requestedStartDate: LocalDate): ActivityMigrateRequest = ActivityMigrateRequest(
+    programServiceCode = programCode,
+    prisonCode = prisonId,
+    startDate = maxOf(startDate, requestedStartDate),
+    endDate = endDate,
+    capacity = max(1, capacity),
+    description = description,
+    payPerSession = ActivityMigrateRequest.PayPerSession.valueOf(payPerSession),
+    runsOnBankHoliday = !excludeBankHolidays,
+    dpsLocationId = internalLocationId.toDpsLocationId(),
```
mikehalmamoj, line 163: The main change is to get the DPS location from the mapping service and stop sending the NOMIS location details. This will allow activities to remove internal location from the API and make dps location non-nullable.

petergphillips (approved):
LGTM

PR #1158, opened 2025-03-05
sdit-2644 sentence migration and move to new model

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/Transformer.kt
```diff
@@ -14,30 +14,35 @@ import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.courtsentencing.m
...
   appearances = this.courtEvents.map { ca ->
-    ca.toMigrationDpsCourtAppearance()
+    ca.toMigrationDpsCourtAppearance(this.sentences)
   },
   courtCaseLegacyData = CourtCaseLegacyData(
     caseReferences = this.caseInfoNumbers.map {
       CaseReferenceLegacyData(
         offenderCaseReference = it.reference,
-        updatedDate = LocalDateTime.parse(it.createDateTime),
+        updatedDate = LocalDateTime.parse(it.createDateTime.toString()),
```
petergphillips, line 45: can't tell what `createDateTime` is coming in as, but maybe 
```suggestion
        updatedDate = it.createDateTime,
```
would be simpler?

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/Transformer.kt
lbennett-moj: whoops yes 😆

PR #1161, opened 2025-03-06
SDIT-2181: ✨ Get gradle to download json and add check docs bash script

petergphillips:


build.gradle.kts
```diff
@@ -59,132 +63,113 @@ kotlin {
...
     packageName = "nomismappings",
+    url = "https://nomis-sync-prisoner-mapping-dev.hmpps.service.justice.gov.uk/v3/api-docs",
   ),
-  // https://prison-person-api-dev.prison.service.justice.gov.uk/v3/api-docs
   ModelConfiguration(
-    name = "buildPrisonPersonApiModel",
-    input = "prison-person-api-docs.json",
-    output = "prisonperson",
+    name = "prison-person",
     packageName = "prisonperson",
```
petergphillips, line 146: haven't specified a url here as there is no API to call

build.gradle.kts
```diff
@@ -59,132 +63,113 @@ kotlin {
...
     packageName = "locations",
+    url = "https://locations-inside-prison-api-dev.hmpps.service.justice.gov.uk/v3/api-docs",
   ),
-  // https://nomis-prisoner-api-dev.prison.service.justice.gov.uk/v3/api-docs
   ModelConfiguration(
-    name = "buildNomisSyncApiModel",
-    input = "nomis-sync-api-docs.json",
-    output = "nomissync",
+    name = "nomis-sync",
     packageName = "nomissync",
```
petergphillips, line 132: haven't specified a url as will be removed once we move to nomis-prisoner

README.md
```diff
@@ -61,42 +61,19 @@ Then run any of the `bash` scripts at the root of this project to send events to
...
-curl https://hmpps-person-record-dev.hmpps.service.justice.gov.uk/v3/api-docs | jq . > openapi-specs/core-person-api-docs.json
+In the build.gradle.kts there is a `ModelConfiguration` for each of the models.  This contains the URL to be used to
+obtain the JSON Open API specification and also the package name to be used for the generated models.  For each model
+configuration two tasks are created:
+1. `write<ModelName>Json` - this task will download the Open API specification and save it to `openapi-specs`.
+2. `build<ModelName>ApiModel` - this task will generate the models from the specification
+
+So, for example, running
+```shell
+./gradlew clean writeNonAssociationsJson compileKotlin compileTestKotlin
```
mikehalmamoj, line 72: Different to nomis-update which has `./gradlew clean writeNonAssociationsJson buildNonAssociationsApiModel compileKotlin compileTestKotlin`. Is that on purpose?

mikehalmamoj: Also thinking about it I'm not sure you need to clean - that forces all models to be rebuilt. The build model task should have an input which is the json, and therefore will only run if the json has changed.

check-api-docs.bash
```diff
@@ -0,0 +1,48 @@
...
+  OUTPUT_COMPARISON=true
+else
+  OUTPUT_COMPARISON=false
+fi
+
+OLD_VERSION=$(jq -r .info.version "$SPECS_JSON")
+echo "old_version=$OLD_VERSION" >>"$GITHUB_OUTPUT"
+
+# build the model and grab new json
+./gradlew clean "$MODEL_TASK" "$JSON_TASK"
```
mikehalmamoj, line 20: Again not sure about clean here

mikehalmamoj (approved):

PR #1222, opened 2025-03-20
SDIT-2609: ✨ Add new refresh for completed migrations endpoint

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/visitbalances/VisitBalanceMigrationResource.kt
```diff
@@ -171,4 +171,44 @@ class VisitBalanceMigrationResource(
     ],
   )
   suspend fun activeMigration() = migrationHistoryService.getActiveMigrationDetails(MigrationType.VISIT_BALANCE)
+
+  @PostMapping("/{migrationId}/refresh")
```
petergphillips, line 175: it does seem weird that this isn't in a generic migration resource rather than having to be added for each migration separately.

mikehalmamoj: I suspect there's no strategic reason for this - there's still a lot of stuff in this project (and the UI) which is generic and could be consolidated.

mikehalmamoj (approved):

PR #1230, opened 2025-03-21
SDIT-2375 Process move bookings for contacts

andymarke:
* Reset contacts in DPS to match NOMIS after the latest booking has moved
* When the latest booking is not active also reset the other prisoner


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/alerts/AlertsSynchronisationService.kt
```diff
@@ -281,15 +283,6 @@ class AlertsSynchronisationService(
     }
   }
 
-  enum class WhichMoveBookingPrisoner {
```
andymarke, line 284: move to util file for reuse

petergphillips (approved):
LGTM

PR #1250, opened 2025-03-27
SDIT-2680: ♻️ Remove unused endpoints now that there are generic ones

petergphillips:


.circleci/config.yml
```diff
@@ -7,10 +7,10 @@ parameters:
     default: syscon-alerts-security
   test-executor-packages:
     type: string
-    default: "activities alerts appointments courtsentencing csip incidents locations sentencing visits personalrelationships casenotes organisations"
+    default: "activities alerts appointments courtsentencing csip incidents locations sentencing visits personalrelationships casenotes organisations coreperson visitbalances"
```
petergphillips, line 10: one of the parallel tests was taking 7 minutes, turns out that each time a new migration is added they need to be added here so that can still be split

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/resources/MigrationHistoryResource.kt
```diff
@@ -1,117 +0,0 @@
...
-import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.config.ErrorResponse
-import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.persistence.repository.MigrationHistory
-import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.HistoryFilter
-import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.MigrationHistoryService
-import java.time.LocalDateTime
-
-@RestController
-@RequestMapping("/", produces = [MediaType.APPLICATION_JSON_VALUE])
-class MigrationHistoryResource(private val migrationHistoryService: MigrationHistoryService) {
-  @PreAuthorize("hasRole('ROLE_MIGRATION_ADMIN')")
```
petergphillips, line 28: these two look like they were only added for the tests - have called the repository instead to `deleteAll`.

steverendell (approved):
lgtm

PR #1274, opened 2025-04-04
SDIT-2694 Basic plumbing for merge court support

andymarke:
Half working version based on existing DPS endpoints that shoud at least:
* Show new cases being created
* Show court cases and sentences being deactivated

However, without new endpoint cases from the removed prisoner number will not see the cases moved to the retained prisoner.

Done
* Call existing (unsuitable) DPS endpoint

TODO
* call new DPS endpoint to sync
* call mapping for new mappings


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingSynchronisationService.kt
```diff
@@ -779,11 +780,21 @@ class CourtSentencingSynchronisationService(
...
     event: OffenderSentenceEvent,
     consecSequence: Int,
-  ): String = mappingApiService.getSentenceOrNullByNomisId(
+  ): String = getConsecutiveSequenceMappingOrThrow(
     bookingId = event.bookingId,
+    sentenceSequence = event.sentenceSeq,
+    consecSequence = consecSequence,
+  )
+
+  private suspend fun getConsecutiveSequenceMappingOrThrow(
```
andymarke, line 789: small defector so it can be called without a prisoner event in context

mikehalmamoj (approved):

PR #1278, opened 2025-04-07
SDIT-2694 Basic plumbing for merge court support

andymarke:
Half working version based on existing DPS endpoints that should at least:
* Show new cases being created
* Show court cases and sentences being deactivated

However, without new endpoint cases from the removed prisoner number will not see the cases moved to the retained prisoner.

Done
* Call existing (unsuitable) DPS endpoint
* call mapping for new mappings TODO
* call new DPS endpoint to sync


steverendell (approved):
lgtm

PR #1294, opened 2025-04-11
SDIT-2713 Migrate linked court cases

andymarke:
A linked court cases is where multiple cases are heard at the same time so the charges from one are copied to another. The original cases is deactivated in NOMIS

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingMigrationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingMigrationService.kt
@@ -21,7 +21,6 @@ import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomismappings.mod
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomismappings.model.CourtCaseMigrationMappingDto
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomismappings.model.CourtChargeMappingDto
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomismappings.model.SentenceMappingDto
-import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomisprisoner.model.CourtCaseResponse
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomisprisoner.model.PrisonerId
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingMigrationService.kt
```diff
@@ -90,31 +89,6 @@ class CourtSentencingMigrationService(
       }
   }
 
-  // temp until DPS handles linked cases
```
andymarke, line 93: main change, no longer ignore these case but use the DPS rules to link the cases

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/Transformer.kt
```diff
@@ -47,7 +48,13 @@ fun CourtCaseResponse.toMigrationDpsCourtCase() = MigrationCreateCourtCase(
     },
   ),
   active = this.caseStatus.code == "A",
-
```
andymarke, line 50: DPS has 3 states that represent:

- case/charge not involved in any case linking
- is involved and is the source; so is read-only therefore merged = true
- is involved and is the target; so not read-only  therefore merged = false

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingMigrationIntTest.kt
```diff
@@ -178,114 +181,35 @@ class CourtSentencingMigrationIntTest(
 
       webTestClient.performMigration()
 
-      await untilAsserted {
```
andymarke, line 181: just converted this to assert on the object sent for clarity

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingMigrationIntTest.kt
```diff
@@ -305,117 +229,149 @@ class CourtSentencingMigrationIntTest(
...
```
andymarke, line 308: just converted this to assert on the object sent for clarity

petergphillips (approved):
LGTM

PR #1344, opened 2025-05-01
SDIT-2723 Allow large DPS timeout when running manual repair

andymarke:
Repair endpoint for certain POM contacts can take over 30 seconds - this ensures we wait a longer time given this is only called manually


steverendell (approved):
lgtm

PR #1375, opened 2025-05-09
SDIT-2761 Handle 409 for DPS create contact

andymarke:
Handle by doing nothing and just emitting telemetry that we can double check since it should mean we have double POSTed due to network error


petergphillips (approved):
LGTM

PR #1385, opened 2025-05-15
SDIT-2769 Ignore contacts CRUD events for prisoner to prisoner contacts

andymarke:
DPS does not support this tiny number of prisoner to prisoner contacts (7 in last  year)


petergphillips (approved):
LGTM

PR #1404, opened 2025-05-21
SDIT-2761 Retry on duplicate contacts

andymarke:
When a contact is being removed and added at the same time, DPS will correctly return 409 if the contact delete hasn't be processed yet. So just keep retrying until there is manual intervention.


steverendell (approved):
lgtm

PR #1421, opened 2025-06-03
SDIT-2805 No longer retry just mappings

andymarke:
Try to fix issue where POST to DPS fails so mappings are stale - so retry the whole booking contact resync


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/personalrelationships/ContactPersonSynchronisationService.kt
```diff
@@ -1228,26 +1227,16 @@ class ContactPersonSynchronisationService(
...
     telemetry: Map<String, Any>,
   ) {
-    try {
-      mappingApiService.replaceMappingsForPrisoner(offenderNo, mapping)
-      telemetryClient.trackEvent(
-        "from-nomis-synch-contactperson-booking-changed",
-        telemetry,
-      )
-    } catch (e: Exception) {
-      log.error("Failed to replace prisoner person for $mapping", e)
```
andymarke, line 1243: don't catch exception, just retry the whole event which is idempotent anyway

mikehalmamoj (approved):

PR #1437, opened 2025-06-06
SDIT-2768 OFFENDER_FIXED_TERM_RECALLS-UPDATED integration

andymarke:
Dropped INSERT and DELETED

INSERT happens when a sentence is inserted - so the RTC date is sent then anyway DELETE never happens, there is no way to delete a date in NOMIS

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingEventListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingEventListener.kt
@@ -62,8 +62,7 @@ class CourtSentencingEventListener(
               "OFFENDER_CASE_IDENTIFIERS-UPDATED",
               -> courtSentencingSynchronisationService.nomisCaseIdentifiersUpdated(eventType, sqsMessage.Message.fromJson())
               "prison-offender-events.prisoner.merged" -> courtSentencingSynchronisationService.prisonerMerged(sqsMessage.Message.fromJson())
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingEventListener.kt
```diff
@@ -62,8 +62,7 @@ class CourtSentencingEventListener(
               "OFFENDER_CASE_IDENTIFIERS-UPDATED",
               -> courtSentencingSynchronisationService.nomisCaseIdentifiersUpdated(eventType, sqsMessage.Message.fromJson())
               "prison-offender-events.prisoner.merged" -> courtSentencingSynchronisationService.prisonerMerged(sqsMessage.Message.fromJson())
-              "OFFENDER_FIXED_TERM_RECALLS-INSERTED", "OFFENDER_FIXED_TERM_RECALLS-UPDATED", "OFFENDER_FIXED_TERM_RECALLS-DELETED" -> courtSentencingSynchronisationService.nomisRecallReturnToCustodyDataChanged(sqsMessage.Message.fromJson())
-
+              "OFFENDER_FIXED_TERM_RECALLS-UPDATED" -> courtSentencingSynchronisationService.nomisRecallReturnToCustodyDataChanged(sqsMessage.Message.fromJson())
               else -> log.info("Received a message I wasn't expecting {}", eventType)
```
andymarke, line 66: deletes can't happen in NOMIS and inserts happening in the same transaction as the sentence insert so no point in updating DPS with the same data

karenmillermoj (approved):
lgtm

PR #1496, opened 2025-06-27
SDIT-2909 Fixed case linking logic

andymarke:
And removed DPS fields due to be deleted

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/Transformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/Transformer.kt
@@ -33,11 +33,10 @@ import java.time.LocalDate
 import java.time.LocalDateTime
 import java.util.UUID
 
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/Transformer.kt
```diff
@@ -145,41 +142,23 @@ fun CourtEventChargeResponse.toDpsCharge() = LegacyUpdateCharge(
...
+    postedDate = LocalDate.now().toString(),
+    outcomeDescription = this.resultCode1?.description,
+    nomisOutcomeCode = this.resultCode1?.code,
+    outcomeDispositionCode = this.resultCode1?.dispositionCode,
+    outcomeConvictionFlag = this.resultCode1?.conviction,
+  ),
+  offenceEndDate = this.offenceEndDate,
+  chargeNOMISId = chargeId,
+  sentence = dpsSentence,
+  mergedFromCaseId = linkedCaseDetails?.caseId,
```
andymarke, line 159: only 2 fields now need by DPS - so much simplier

PR #1499, opened 2025-06-30
SDIT-2889 Convert person migration to prisoner restrictions

andymarke:
TODO:

* Get prisoner number back from NOMIS API
* Fix tests that rely of prisoner number


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/personalrelationships/ContactPersonMigrationMessageListener.kt
```diff
@@ -20,7 +20,7 @@ import java.util.concurrent.CompletableFuture
 class ContactPersonMigrationMessageListener(
```
andymarke, line 20: convert this to use PrisonerRestrictions rather than deleting and created from new. This will be a one shot migration so happy to cheat a little bit here

PR #1505, opened 2025-07-01
SDIT-2919:✨ Ignore any NOMIS incident events if DPS is in charge

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/incidents/IncidentsSynchronisationService.kt
```diff
@@ -108,15 +112,21 @@ class IncidentsSynchronisationService(
   }
 
   suspend fun synchroniseIncidentUpdate(event: IncidentsOffenderEvent) {
-    // Should never happen
+    // two way sync so don't want to send an event to DPS that originated there
```
karenmillermoj, line 115: Once the audit module name is added we may also need to check for the 6am-6:30am issue where it occasionally doesn't get set too

karenmillermoj (approved):
lgtm

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/incidents/IncidentsSynchronisationService.kt
petergphillips: agreed, but at present it is always `null` so will mean we would send anything to DPS 😱

PR #1514, opened 2025-07-02
SDIT-2889 Log IDs for duplicate

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/personalrelationships/PrisonerRestrictionMigrationService.kt
```diff
@@ -99,6 +99,9 @@ class PrisonerRestrictionMigrationService(
         telemetryClient.trackEvent(
           "nomis-migration-contactperson-duplicate",
           mapOf(
+            "offenderNo" to mapping.offenderNo,
+            "dpsId" to mapping.dpsId,
+            "nomisId" to mapping.nomisId,
             "duplicateDpsId" to duplicateErrorDetails.duplicate.dpsId,
```
andymarke, line 105: these should also be logged from below if mapping service does it job - but log here also just in case

steverendell (approved):
lgtm

PR #1571, opened 2025-07-22
SDIT-2948-Add-from-nomis-event-handling-for-insert-events

steverendell:
Standard from-nomis insert sync logic, but with some differences.
In Nomis there is a list of 'offender transactions' (multiple rows) with a particular transaction id, each of which has a list of, or multiple 'general ledger transaction' rows. Or alternatively there is just a list of 'general ledger transaction's. 
On insert to either of these tables, we first get all the Nomis data so far and if this includes  'offender transactions'  we call a DPS upsert for these along with their GL child lists. If it does not, we call an alternative GL upsert endpoint with just these GLs.
The first event will result in the creation of the transaction id in DPS and a mapping, the rest will not.


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/finance/TransactionSynchronisationIntTest.kt
```diff
@@ -0,0 +1,829 @@
...
+  @DisplayName("OFFENDER_TRANSACTIONS-INSERTED")
+  inner class TransactionInserted {
+
+    @Nested
+    @DisplayName("When transaction was created in DPS")
+    inner class DPSCreated {
+
+      @BeforeEach
+      fun setUp() {
+        awsSqsFinanceOffenderEventsClient.sendMessage(
```
andymarke, line 71: You can just do financeOffenderEventsQueue.sendMessage

andymarke (approved):
LGTM

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/finance/TransactionSynchronisationIntTest.kt
steverendell: ok will change

PR #1572, opened 2025-07-22
SDIT-2517 Don't use the NOMIS appointments OpenApi generated classes as they're different

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/appointments/AppointmentsMigrationService.kt
```diff
@@ -163,3 +160,26 @@ class AppointmentsMigrationService(
...
+  val startDateTime: LocalDateTime? = null,
+  val endDateTime: LocalDateTime? = null,
+  val comment: String? = null,
+  val subtype: String,
+  val status: String,
+  val createdDate: LocalDateTime,
+  val createdBy: String,
+  val modifiedDate: LocalDateTime? = null,
+  val modifiedBy: String? = null,
+)
```
mikehalmamoj, line 181: I previously replaced this with the swagger generated model without comparing the two - that was a mistake 😢

PR #1582, opened 2025-07-23
SDIT-2956 Add restrictions repair endpoint

andymarke:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/personalrelationships/ContactPersonDataRepairResource.kt
```diff
@@ -60,7 +60,7 @@ class ContactPersonDataRepairResource(
     }
   }
 
-  @PostMapping("/prisoners/{offenderNo}/resynchronise")
+  @PostMapping("/prisoners/{offenderNo}/contacts/resynchronise")
   @ResponseStatus(HttpStatus.OK)
```
andymarke, line 64: original name didn't reference contacts at all - I am an idiot

PR #1584, opened 2025-07-24
SDIT-2243 Fix retry with duplicate adjustment bug

andymarke:
In the extreme scenario that mapping fails then fails on retry with 409 the custom event was not tracked


petergphillips (approved):
LGTM

PR #1601, opened 2025-07-29
SDIT-2977 Flaky test fix

andymarke:
The issue was errorVisibilityTimeout was 1 second running tests, so occasionally we hit retries sometimes we didn't

So changed about 30% tests to waitForCompletion correctly and removed the await assert which was no longer needed.

Still only 60% of tests to do


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingSynchronisationIntTest.kt
```diff
@@ -101,14 +102,13 @@ class CourtSentencingSynchronisationIntTest : SqsIntegrationTestBase() {
...
-        awsSqsCourtSentencingOffenderEventsClient.sendMessage(
-          courtSentencingQueueOffenderEventsUrl,
+        courtSentencingOffenderEventsQueue.sendMessage(
           courtCaseEvent(
             eventType = "OFFENDER_CASES-INSERTED",
             auditModule = "DPS_SYNCHRONISATION",
           ),
         ).also {
-          waitForTelemetry()
+          waitForAnyProcessingToComplete()
```
andymarke, line 111: just being consistent with other projects using waitForAnyProcessingToComplete

petergphillips (approved):
LGTM

PR #1619, opened 2025-07-31
SDIT-2874 Add check and save mappings

mikehalmamoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/ExternalMovementsMigrationIntTest.kt
```diff
@@ -60,6 +72,77 @@ class ExternalMovementsMigrationIntTest(
...
+          assertThat(it["migrationId"]).isEqualTo(migrationId)
+        },
+        isNull(),
+      )
+    }
+  }
+
+  @Nested
+  inner class MappingErrorRecovery {
+    @BeforeEach
```
mikehalmamoj, line 113: Also need to handle mapping already exists and duplicate mapping on create - coming soon

PR #1660, opened 2025-08-14
SDIT-2873 Update to match real create migration mapping endpoint

mikehalmamoj:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/wiremock/MappingApiMockServer.kt
```diff
@@ -658,9 +661,9 @@ class MappingApiMockServer : WireMockServer(WIREMOCK_PORT) {
     )
   }
 
-  fun stubMappingCreateFailureFollowedBySuccess(url: String) {
+  fun stubMappingCreateFailureFollowedBySuccess(url: String, method: (UrlPattern) -> MappingBuilder = WireMock::post) {
```
mikehalmamoj, line 664: Add http method as parameter so I can re-use this for my put endpoint

PR #1692, opened 2025-08-27
SDIT-2983 Call DPS booking create endpoint for clone

andymarke:
Also call getCourtCases NOMIS API endpoint.

The DPS endpoint is currently identical to the migration endpoint other than the type names so a lot of copy and paste from the migration code.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/Transformer.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/Transformer.kt
@@ -1,5 +1,12 @@
 package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.courtsentencing
 
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.courtsentencing.model.BookingCreateCharge
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.courtsentencing.model.BookingCreateCourtAppearance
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/Transformer.kt
```diff
@@ -56,6 +63,29 @@ fun CourtCaseResponse.toMigrationDpsCourtCase() = MigrationCreateCourtCase(
     null
   },
 )
+fun CourtCaseResponse.toBookingCloneDpsCourtCase() = BookingCreateCourtCase(
```
andymarke, line 66: All these are identical to their migration counter parts - not much we can do about the duplication given the DTOs are generated from DPS OpenAPI spec - so they are just a bunch of copy and paste from the the code above in each case

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingMigrationIntTest.kt
```diff
@@ -676,6 +683,66 @@ fun dpsMigrationCreateResponseWithTwoAppearancesAndTwoCharges(): MigrationCreate
     sentenceTerms = sentenceTermIds,
   )
 }
+fun dpsBookingCloneCreateResponseWithTwoAppearancesAndTwoCharges(): BookingCreateCourtCasesResponse {
```
andymarke, line 686: Same concept for this test data - identical to existing migration response other than the types

petergphillips (approved):
LGTM

PR #1725, opened 2025-09-04
SDIT-3025 Add caching to transaction sync from nomis

steverendell:
Limit duplicate sync calls to DPS by introducing a grace period


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/finance/TransactionEventListener.kt
```diff
@@ -36,16 +37,25 @@ class TransactionEventListener(
...
                 sqsMessage.Message.fromJson(),
                 messageId,
               )
               // NOTE: Several inserts per second!
+              /*
+              when event arrives, check DB table of txn_ids:
+               if not there, add and requeue message for 5s time
+               if it is there and target time not yet arrived (or not expired), ignore
+              when requeued message arrives: remove from cache if necessary and call api
+               */
```
steverendell, line 50: Actually this comment isnt entirely accurate now Im checing it - there is no stored 'target time'

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/finance/TransactionSynchronisationService.kt
```diff
@@ -23,17 +26,44 @@ class TransactionSynchronisationService(
...
+    }
+
+    if (transactionIdBufferRepository.existsById(event.transactionId)) {
+      log.debug("Skipping synchronisation of transaction ${event.transactionId} as id already recorded")
+      return
+    }
+
+    if (transactionIdBufferRepository.addId(event.transactionId) == 0) {
+      log.debug("Skipping synchronisation of transaction ${event.transactionId} as failed to record id")
+      return
```
andymarke, line 50: Shouldn't this be an exception?

steverendell: This is the race condition where 2 threads try to insert at the same time and this is the one that fails. I think other errors would result in an exception from the repository method but ill double check ..

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/SynchronisationQueueService.kt
```diff
@@ -29,27 +29,38 @@ class SynchronisationQueueService(
...
+    val builder = SendMessageRequest.builder()
+      .queueUrl(queue.queueUrl)
+      .messageBody(sqsMessage.toJson())
+      .eventTypeMessageAttributes("prisoner-from-nomis-synchronisation-$messageType")
+
+    if (delaySeconds > 0) {
+      builder.delaySeconds(delaySeconds)
     }
+
+    queue.sqsClient.sendMessage(builder.build())
```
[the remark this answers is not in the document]
andymarke: Have a look at https://github.com/ministryofjustice/hmpps-prisoner-to-nomis-update/blob/d213c4994d16bb3353068e74e836756de4ccdc17/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonertonomisupdate/services/QueueService.kt

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/finance/TransactionSynchronisationService.kt
```diff
@@ -23,17 +26,44 @@ class TransactionSynchronisationService(
...
+    queueService.sendMessage(
+      messageType = SynchronisationMessageType.PERFORM_TRANSACTION_SYNC.name,
+      synchronisationType = SynchronisationType.FINANCE,
+      message = EncapsulatedTransaction(event, requestId),
+      telemetryAttributes = mapOf("transactionId" to event.transactionId.toString()),
+      delaySeconds = forwardingDelaySeconds,
+    )
+  }
+
+  suspend fun transactionInserted(encapsulatedTransaction: InternalMessage<EncapsulatedTransaction>) {
```
andymarke, line 62: So when you have the hundreds of transactions scenario this will also be called hundred times all at the same time - but 5 seconds after the hundreds of xtag events are received?

src/test/resources/application-test.yml
```diff
@@ -60,6 +60,8 @@ courtsentencing:
   page:
     size: 10
   has-migrated-data: true
+finance:
+  transactions.forwardingDelaySeconds: 1
```
andymarke, line 64: you might want to consider moving this to 0 - else I guess every test will take at least 1 second

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/SynchronisationQueueService.kt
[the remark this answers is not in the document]
steverendell: Good point - I think the best thing might be to catch any exception from the send, delete the transaction id from the DB and rethrow. Ill check that send in the events repo to ensure errors are indeed detected.

andymarke (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/finance/TransactionSynchronisationService.kt
steverendell: No, this is only called after the initial event successfully inserted the transaction id into the db table, and then queued the 5s delay message

src/test/resources/application-test.yml
steverendell: One problem is that the multi-event test can fail so Ive separated that into another test class so i can set its forwardingDelaySeconds back to 1.

PR #1733, opened 2025-09-08
SDIT-2845 Check parent application exist before trying to create children

mikehalmamoj:
Attempting to create either an outside movement or scheduled movement in DPS BEFORE the movement application has been created will fail - DPS will have a FK missing for the application. (This can happen if events are processed out of order).

In this situation we just fail the event so it goes back to the DLQ and it should work on a retry.


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/ExternalMovementsSyncService.kt
```diff
@@ -130,6 +132,9 @@ class ExternalMovementsSyncService(
       }
   }
 
+  private suspend fun requireParentApplicationExists(nomisApplicationId: Long) = mappingApiService.getApplicationMapping(nomisApplicationId)
+    ?: throw ParentEntityNotFoundRetry("Application $nomisApplicationId not created yet so children cannot be processed")
+
```
mikehalmamoj, line 137: This is where we fail if the parent application has not been created yet. The exception should be caught, its message added to telemetry, and then the exception re-thrown to put the message onto the DLQ. The error handling is currently done in the `track` helper function.

PR #1829, opened 2025-10-06
SDIT-3078 Repair prisoner court cases endpoint

andymarke:
Essentially a wrapper for migration without the full migration excesses

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingRepairService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingRepairService.kt
@@ -1,0 +1,38 @@
+package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.courtsentencing
+
+import com.microsoft.applicationinsights.TelemetryClient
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingRepairService.kt
```diff
@@ -0,0 +1,38 @@
...
+  val courtSentencingMigrationService: CourtSentencingMigrationService,
+  val telemetryClient: TelemetryClient,
+) {
+  suspend fun resynchronisePrisonerCourtCases(offenderNo: String) {
+    courtSentencingMigrationService.synchronisePrisonerCases(
+      offenderNo = offenderNo,
+      deleteExisting = true,
+      context = MigrationContext(
+        type = MigrationType.COURT_SENTENCING,
+        migrationId = LocalDateTime.now().toString(),
```
andymarke, line 22: a pragmatic compromise here - pretend there is a migration ID so we don't have to make big changes around the retry mapping functionality that assumes we have a migrationId just so it can log it

mikehalmamoj (approved):

PR #2685, opened 2026-04-13
SDIT-3671 Handle official visit converting tp social visit

andymarke:
When this happens we wil get an update visit event but the call to NOMIS API will fail with a 400 error and have a specific error message

When this happens treat it as an official visit delete

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/officialvisits/OfficialVisitsNomisApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/officialvisits/OfficialVisitsNomisApiService.kt
@@ -4,6 +4,8 @@ import kotlinx.coroutines.reactor.awaitSingle
 import org.springframework.beans.factory.annotation.Qualifier
 import org.springframework.stereotype.Service
 import org.springframework.web.reactive.function.client.WebClient
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.helpers.SuccessOrBadRequest
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.helpers.awaitSuccessOrBadRequestErrorMessage
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomisprisoner.api.OfficialVisitsResourceApi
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/officialvisits/OfficialVisitsSynchronisationService.kt
```diff
@@ -94,20 +95,44 @@ class OfficialVisitsSynchronisationService(
...
-            dpsApiService.updateVisit(mapping.dpsId.toLong(), nomisVisit.toSyncUpdateOfficialVisitRequest())
+          val nomisVisitResult = nomisApiService.getOfficialVisitOrBadRequestErrorMessage(event.visitId)
+          if (nomisVisitResult.isError) {
+            nomisVisitResult.assertNoLongerOfficialVisitOrThrow()
+            officialVisitConvertedToSocialVisit(event, mapping.dpsId)
+          } else {
+            dpsApiService.updateVisit(
+              mapping.dpsId.toLong(),
+              nomisVisitResult.successResponse!!.toSyncUpdateOfficialVisitRequest(),
+            )
```
mikehalmamoj, line 109: It's a little bit confusing that the happy path is buried in an if then an else, might be easier if that was the first thing you see, or maybe the non-happy path is more defensive and they return.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/officialvisits/OfficialVisitsNomisApiService.kt
```diff
@@ -48,6 +50,16 @@ class OfficialVisitsNomisApiService(@Qualifier("nomisApiWebClient") private val
...
 
+  suspend fun getOfficialVisitOrBadRequestErrorMessage(
+    visitId: Long,
+  ): SuccessOrBadRequest<OfficialVisitResponse> = with(api) {
+    awaitSuccessOrBadRequestErrorMessage(
+      getOfficialVisitRequestConfig(
+        visitId = visitId,
+      ),
+    )
+  }
```
mikehalmamoj, line 61: Look at you

mikehalmamoj (approved):

PR #2687, opened 2026-04-15
SDIT-3668: 🐛 Fix api docs script as dependency order corrected

petergphillips:


check-api-docs.bash
```diff
@@ -19,11 +19,14 @@ export _JAVA_OPTIONS="-Xmx768m -XX:ParallelGCThreads=2 -XX:ConcGCThreads=2 -XX:P
 OLD_VERSION=$(jq -r .info.version "$SPECS_JSON")
 echo "old_version=$OLD_VERSION" >>"$GITHUB_OUTPUT"
 
-# build the model and grab new json
-./gradlew "$MODEL_TASK" "$JSON_TASK"
+# build the model
+./gradlew "$MODEL_TASK"
```
petergphillips, line 23: this previously worked as expected because there was no dependency order between the two tasks so the build would happen and afterwards the new json would be written. At this stage we just want to build the model.

check-api-docs.bash
```diff
@@ -33,11 +36,8 @@ if [[ "$OLD_VERSION" == "$NEW_VERSION" ]]; then
   exit 0
 fi
 
-# build the model again
-./gradlew "$MODEL_TASK"
-
 # and compare
-if ! diff -r build/model_copy "${WORKING_DIR}/build/generated/$PROJECT" > "${WORKING_DIR}/build/api.diff"; then
+if ! diff -r "${WORKING_DIR}/build/model_copy" "${WORKING_DIR}/build/generated/$PROJECT" > "${WORKING_DIR}/build/api.diff"; then
```
karenmillermoj, line 40: so was WORKING_DIR empty before? otherwise this wouldn't have worked? as the earlier mv moves it to ${WORKING_DIR}/build/model_copy

karenmillermoj (approved):
lgtm

check-api-docs.bash
petergphillips: sorry, this change makes no difference in this project as both are the same. I fixed the script in `prisoner-search` first and copied the script from there. It already had the change as for that project `build/model_copy`  is different (due to sub projects).

PR #2702, opened 2026-04-16
SDIT-2832 Copy migration into taps package

mikehalmamoj:
For TAPs, we've repurposed the migration as a repair utility so we need to keep it. So it needs moving into the TAP package and we'll write another migration for the next entity we tackle. The only change is to the endpoint URLs - changed to indicate they apply only to taps.

Still keeping the old migration endpoints for now, until I've had a chance to switch to the new ones in the migration UI.


petergphillips (approved):
LGTM

PR #2714, opened 2026-04-20
SDIT-2832 Switch get all offender taps to new NOMIS API

mikehalmamoj:
The only real change is to switch to the new endpoint to get all offender TAP details. However the model returned by that endpoint is huge and so this change is huge.

Ultimately we'll have to rely on the tests here.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/taps/TapsNomisApiService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/taps/TapsNomisApiService.kt
@@ -1,11 +1,12 @@
-package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.movements
+package uk.gov.justice.digital.hmpps.prisonerfromnomismigration.movements.taps
 
 import org.springframework.beans.factory.annotation.Qualifier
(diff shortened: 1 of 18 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/taps/TapsNomisApiService.kt
```diff
@@ -1,24 +1,23 @@
...
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomisprisoner.model.TemporaryAbsenceResponse
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomisprisoner.model.TemporaryAbsenceReturnResponse
 
 @Service
-class ExternalMovementsNomisApiService(@Qualifier("nomisApiWebClient") private val webClient: WebClient) {
-  suspend fun getTemporaryAbsencesOrNull(offenderNo: String): OffenderTemporaryAbsencesResponse? = webClient.get()
-    .uri {
-      it.path("/movements/{offenderNo}/temporary-absences")
-        .build(offenderNo)
-    }
```
mikehalmamoj, line 21: This is the only material code change - switch to the new API and the new model.

petergphillips (approved):
LGTM

PR #2720, opened 2026-04-21
SDIT-2832 Move TAP specific services

mikehalmamoj:
This refactor consists mainly of moving an renaming TAP specific services, plus some tidying of imports and ktlint messed around with some stuff.


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/taps/TapApplicationIntTest.kt
```diff
@@ -35,7 +35,7 @@ import java.time.Duration
 import java.time.LocalDate
 import java.time.LocalDateTime
 import java.time.LocalTime
-import java.util.UUID
+import java.util.*
```
karenmillermoj, line 38: I'm never sure about import with .* - ktlint hasn't picked this up for awhile so possibly ok now - but can see you did revert this back to java.util.UUID on another file

karenmillermoj (approved):
lgtm

PR #2721, opened 2026-04-21
SDIT-2832 Move TAP specific services

mikehalmamoj:
Mostly moving and renaming TAP services


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/ExternalMovementsMovementRouter.kt
```diff
@@ -1,21 +0,0 @@
...
-class ExternalMovementsMovementRouter(
-  private val tapMovementService: TapMovementService,
-) {
-  companion object {
-    val log: Logger = LoggerFactory.getLogger(this::class.java)
-  }
-
-  suspend fun externalMovementChanged(event: ExternalMovementEvent) = when (event.movementType) {
-    TAP -> tapMovementService.tapMovementChanged(event)
-    else -> log.info("Ignoring external movement changed event with type ${event.movementType}, inserted=${event.recordInserted}, deleted=${event.recordDeleted}}")
```
mikehalmamoj, line 19: A proper change here - don't bother with a router to push external movements to the relevant service. Instead we'll have a different queue per sub-domain we're syncing (e.g. one for courts, releases, transfers etc.) This means if we need to action multiple of these sub-domains from a single message they won't be tripping over each other (move booking being a good example).

petergphillips (approved):
LGTM

PR #2784, opened 2026-05-06
SDIT-2832 Experimental - Add support for HTTP/2 DPS APIs

mikehalmamoj:
Spring web client supports HTTP/2 out of the box, but not for unsecured endpoints (e.g. our health checks). java.net.URI doesn't support HTTP/2 but HttpClient does.

As an experiment add HTTP/2 support for external movements health checks and OpenAPI docs to see if it works. Note that we won't find out until the DPS Taps API starts rejecting non HTTP/2 traffic.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -6,6 +6,9 @@ import org.jlleitschuh.gradle.ktlint.tasks.KtLintFormatTask
 import org.openapitools.generator.gradle.plugin.tasks.GenerateTask
 import tools.jackson.databind.json.JsonMapper
 import java.net.URI
+import java.net.http.HttpClient
+import java.net.http.HttpRequest
+import java.net.http.HttpResponse
 import kotlin.io.path.exists
 import kotlin.io.path.pathString
 import kotlin.io.path.Path as KotlinPath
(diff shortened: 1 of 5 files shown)


openapi-specs/movements-api-docs.json
```diff
@@ -1,12 +1,12 @@
...
-  "openapi" : "3.1.0",
+  "openapi" : "3.0.1",
   "info" : {
     "title" : "HMPPS External Movements Api",
     "contact" : {
       "name" : "HMPPS Digital Studio",
       "email" : "feedback@digital.justice.gov.uk"
     },
-    "version" : "2026-04-15.1194.048d4bb"
+    "version" : "2026-05-06.1261.9bccd8b"
```
mikehalmamoj, line 9: Use the updated gradle task to get this spec, so seems to work.

build.gradle.kts
```diff
@@ -86,18 +89,33 @@ abstract class WriteJsonTask : DefaultTask() {
...
+    .version(HttpClient.Version.valueOf(httpVersion))
+    .build()
+    .send(
+      HttpRequest.newBuilder()
+        .uri(URI.create(this))
+        .GET()
+        .build(),
+      HttpResponse.BodyHandlers.ofString(),
+    )
+    .body()
```
mikehalmamoj, line 118: Wasn't really sure where to stick this and a top level private fun caused problems, so duplicated the function in both tasks.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/taps/TapConfiguration.kt
```diff
@@ -34,8 +37,17 @@ class TapConfiguration(
...
-  fun tapsApiHealthWebClient(builder: WebClient.Builder): WebClient = builder.reactiveHealthWebClient(tapsUrl, healthTimeout)
+  fun tapsApiHealthWebClient(builder: WebClient.Builder): WebClient = builder.baseUrl(tapsUrl)
+    .clientConnector(
+      ReactorClientHttpConnector(
+        HttpClient.create()
+          .protocol(if (tapsHttp2) HttpProtocol.H2 else HttpProtocol.HTTP11)
+          .responseTimeout(healthTimeout),
+      ),
+    )
+    .build()
```
mikehalmamoj, line 50: If this works out OK we can add support for http/2 to the kotlin lib

lbennett-moj (approved):

PR #2786, opened 2026-05-06
SDIT-3718 Handle court schedule OUT updated event

mikehalmamoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/court/CourtSchedulerSyncScheduleService.kt
```diff
@@ -114,7 +114,25 @@ class CourtSchedulerSyncScheduleService(
...
+      return
+    }
+
+    track("${TELEMETRY_PREFIX}-updated", telemetry) {
+      val existingMapping = mappingApi.getCourtScheduleMappingOrNull(eventId)
+        ?.also { telemetry["dpsCourtAppearanceId"] = it.dpsCourtAppearanceId }
+        ?: throw IllegalStateException("No mapping found when handling an update event for court schedule $eventId - hopefully messages are being processed out of order and this event will succeed on a retry once the create event is processed. Otherwise we need to understand why the original create event was never processed.")
+
+      syncScheduleOut(prisonerNumber, eventId, telemetry, existingMapping)
+    }
```
mikehalmamoj, line 135: Pretty much the same as an insert except we pass in the existing mapping with the DPS id, and fail if there is no mapping.

andymarke (approved):

PR #2799, opened 2026-05-08
SDIT-3720 Court movement inserted event - edge cases

mikehalmamoj:
Added the following tests for court movement inserted events:

- Schedule exists (get mapping)
- Schedule mapping does not exist
- Created in dps
- Already created
- Duplicate mapping
- Mapping create fails once and retries

And fixed up the service to support them all.


steverendell (approved):
lgtm

PR #2848, opened 2026-05-19
SDIT-3731 Add court migration message handling

mikehalmamoj:
Convert the current test of a single prisoner to call the /migration endpoint instead of calling the service directly - this exercises the migration queues and handlers.


petergphillips (approved):
LGTM

PR #2849, opened 2026-05-19
SDIT-3731 We don't need a spy bean and it causes problems testing locally

mikehalmamoj:
Since we moved back to v2 of the app insights library some team members have started experiencing 5 second delays on the first test running locally. The test times out and we can't test locally. This appears to be caused by the VPN blocking the telemetry client while the test is running. A working theory is that the real telemetry client tries to call out for some reason, but the vpn injects this 5 second delay.

The simple fix is to stop using a telemetry client spy bean for integration tests, and use a mock bean instead - we don't need a spy bean as we'd never have need to call the real bean anyway.


petergphillips (approved):
LGTM

PR #2857, opened 2026-05-19
SDIT-3733 Switch to better domain URL

mikehalmamoj:
Switch from base URL /court to /court-scheduler both in this service and for downstream nomis-mapping clients.


steverendell (approved):
lgtm

PR #2862, opened 2026-05-20
SDIT-3823: ♻️  Tidy references to provide clear separation

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/MigrationHistoryService.kt
```diff
@@ -18,13 +18,12 @@ import java.time.LocalDateTime
 
 @Service
 class MigrationHistoryService(
-  val migrationHistoryRepository: MigrationHistoryRepository,
+  private val migrationHistoryRepository: MigrationHistoryRepository,
   private val jsonMapper: JsonMapper,
   private val hmppsQueueService: HmppsQueueService,
-  private val generalMappingService: GeneralMappingService,
```
petergphillips, line 24: `GeneralMappingService` has references to all the other services in order to get the migration counts.

The other services have a parent of `MigrationService` which references `MigrationHistoryService`. If the history service then has a reference to `GeneralMappingService` then we end up with a circular dependency so can't pull `MigrationService` and `MigrationHistoryService` into a common sub project.

We therefore move the processing to the resource class to avoid the issue.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
```diff
@@ -15,14 +15,15 @@ import org.springframework.web.reactive.function.client.awaitBodilessEntity
 import org.springframework.web.reactive.function.client.awaitBody
 import reactor.core.publisher.Mono
 import tools.jackson.databind.JsonNode
-import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.appointments.AppointmentIdResponse
```
petergphillips, line 18: this is a common service, but referenced classes from sub projects

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/service/NomisApiService.kt
```diff
@@ -82,26 +85,20 @@ class NomisApiService(@Qualifier("nomisApiWebClient") private val webClient: Web
...
-        .build()
-    }
-    .retrieve()
-    .bodyToMono(typeReference<List<VisitRoomUsageResponse>>())
-    .awaitSingle()
+    prisonIds: List<String>,
+    visitTypes: List<String>,
+    fromDateTime: LocalDateTime?,
+    toDateTime: LocalDateTime?,
+  ): List<VisitRoomCountResponse> = visitResourceApi.getVisitRoomCountsByFilter(
```
petergphillips, line 93: switching to using the api classes allows us to avoid the references to the subprojects in this common class

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/wiremock/NomisApiMockServer.kt
```diff
@@ -762,6 +762,7 @@ private fun appointmentResponse(
 ): String =
   """
 {
+  "eventId": 1,
```
petergphillips, line 765: okay to add here since it is a required field in nomis prisoner

mikehalmamoj (approved):

PR #2867, opened 2026-05-21
sdit-3822 sentence repair endpoint

lbennett-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingRepairResource.kt
```diff
@@ -50,6 +52,35 @@ class CourtSentencingRepairResource(
     )
   }
 
+  @PostMapping("/prisoners/{offenderNo}/booking-id/{bookingId}/court-sentencing/case/{caseId}/sentences/{sentenceSeq}/sentence-level/{sentenceLevel}/sentence-category/{sentenceCategory}/repair")
```
andymarke, line 55: I get why you have added category and level - I guess for logging - feels a bit of a shame you couldn't just change the logging and just take booking-id OR case-id 

Having said that hopefully we would be using this much - so maybe it doesn't matter

andymarke (approved):

PR #2876, opened 2026-05-26
SDIT-3772 Add call to get user detail from Nomis

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/users/UsersNomisApiService.kt
```diff
@@ -0,0 +1,18 @@
...
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomisprisoner.model.UserDetails
+
+@Service
+class UsersNomisApiService(@Qualifier("nomisApiWebClient") webClient: WebClient) {
+  private val api = UserResourceApi(webClient)
+
+  suspend fun getUserDetails(userId: Long): UserDetails = api
+    .prepare(api.getUserRequestConfig(userId))
+    .retrieve()
+    .awaitBody()
```
petergphillips, line 17: if you're just calling `awaitBody` then probably easier to call `getUserRequest` instead

petergphillips (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/users/UsersNomisApiService.kt
karenmillermoj: yes - I missed that one - thanks - will do in next PR

PR #2893, opened 2026-05-28
SDIT-3671 Move breach hearing sync to message handler

andymarke:
Rather than relying on the breach hearing flag (which itself is simply using audit data), use the explicit re-sync SQS message received from the to-nomis sync service. This simplifies the "normal" court appearance handler and also breach hearing handling since we don't need all the "does this case exist yet checks?"

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingEventListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingEventListener.kt
@@ -200,7 +200,6 @@ data class CourtAppearanceEvent(
   val caseId: Long?,
   val offenderIdDisplay: String,
   val bookingId: Long,
-  val isBreachHearing: Boolean,
   override val auditModuleName: String?,
 ) : EventAudited
 
@@ -216,7 +215,6 @@ data class RecallBreachCourtEventCharge(
   val eventId: Long,
   val chargeId: Long,
   val offenderIdDisplay: String,
-  val bookingId: Long,
 )
 
 data class CourtEventChargeLinkingEvent(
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingEventListener.kt
```diff
@@ -200,7 +200,6 @@ data class CourtAppearanceEvent(
   val caseId: Long?,
   val offenderIdDisplay: String,
   val bookingId: Long,
-  val isBreachHearing: Boolean,
   override val auditModuleName: String?,
```
andymarke, line 203: no longer used so will eventual remove from event

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingEventListener.kt
```diff
@@ -216,7 +215,6 @@ data class RecallBreachCourtEventCharge(
   val eventId: Long,
   val chargeId: Long,
   val offenderIdDisplay: String,
-  val bookingId: Long,
```
andymarke, line 219: was only used for logging so dropped since we don't have the bookingId in the SQS message

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingSynchronisationService.kt
```diff
@@ -114,10 +114,9 @@ class CourtSentencingSynchronisationService(
         "nomisCourtAppearanceId" to event.eventId.toString(),
         "offenderNo" to event.offenderIdDisplay,
         "nomisBookingId" to event.bookingId.toString(),
-        "isBreachHearing" to event.isBreachHearing.toString(),
       )
     if (isAppearancePartOfACourtCase(event)) {
-      if (event.originatesInDps && !event.isBreachHearing) {
```
andymarke, line 120: All the breach hearing stuff remove from the handling of court-event-inserterd and moved to its own event hander

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/courtsentencing/CourtSentencingSynchronisationService.kt
```diff
@@ -697,7 +707,7 @@ class CourtSentencingSynchronisationService(
     eventId: Long,
     chargeId: Long,
     offenderNo: String,
-    bookingId: Long,
+    bookingId: Long? = null,
```
andymarke, line 710: only used for logging, so it will just be null for the breach hearing  scenario 🤷🏾‍♂️

mikehalmamoj (approved):

PR #2900, opened 2026-06-01
SDIT-3482: ♻️  Minor tweaks in preparation for sub module creation

petergphillips:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/coreperson/Transformer.kt
```diff
@@ -131,8 +131,8 @@ fun OffenderAddress.buildFullAddress(): String {
   val address = mutableListOf<String>()
 
   // Append "Flat" if there is one
-  if (!flat.isNullOrBlank()) {
-    address.add("Flat ${flat.trim()}")
```
petergphillips, line 135: `flat` comes from nomis prisoner model, which will sit in a `common` sub project.
We'll be referencing in a different module (`coreperson`) so [smart cast is then not performed](https://discuss.kotlinlang.org/t/what-is-the-reason-behind-smart-cast-being-impossible-to-perform-when-referenced-class-is-in-another-module/2201) and we get an error.
The workaround is only to reference `flat` once.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/integration/history/MigrationMapping.kt
```diff
@@ -11,8 +11,8 @@ import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.data.MigrationDet
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.data.PagedModelMigrationDetails
 
 abstract class MigrationMapping<MAPPING : Any>(
-  val domainUrl: String,
-  internal val webClient: WebClient,
+  protected val domainUrl: String,
```
petergphillips, line 14: `MigrationMapping` to be moved to a `common` module, but subclasses still need access to the properties

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/court/CourtSchedulerMigrationService.kt
```diff
@@ -216,7 +216,7 @@ fun OffenderCourtMovementsResponse.toDpsRequest(
             ?.let { "$EXTERNAL_REF_PREFIX$it" },
         ),
         created = AtAndBy(schedule.audit.createDatetime, schedule.audit.createUsername),
-        modified = schedule.audit.modifyDatetime?.let { AtAndBy(schedule.audit.modifyDatetime, schedule.audit.modifyUserId!!) },
+        modified = schedule.audit.modifyDatetime?.let { AtAndBy(it, schedule.audit.modifyUserId!!) },
```
petergphillips, line 219: again, use `it` as smart cast impossible here

petergphillips: seemed nicer than doing an unnecessary `!!` on the `modifyDatetime`

mikehalmamoj (approved):

PR #2907, opened 2026-06-01
SDIT-3482: ♻️  Split out service test so can move into sub projects

petergphillips:


src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/helper/SpringAPIServiceTest.kt
```diff
@@ -12,24 +12,8 @@ import org.springframework.security.config.annotation.web.reactive.EnableWebFlux
 import org.springframework.test.context.ActiveProfiles
 import org.springframework.test.context.BootstrapWith
 import org.springframework.web.reactive.config.EnableWebFlux
-import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.alerts.AlertsDpsApiExtension
```
petergphillips, line 15: I want to move this class into the `common` module so that all the tests can reference and use it, however that isn't possible at present since it depends on all the other modules.

src/test/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/coreperson/CorePersonCprApiMockServer.kt
```diff
@@ -13,6 +13,7 @@ import org.junit.jupiter.api.extension.ExtensionContext
 import org.springframework.http.HttpStatus
 import org.springframework.test.context.junit.jupiter.SpringExtension
 import tools.jackson.databind.json.JsonMapper
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.coreperson.CorePersonCprApiExtension.Companion.jsonMapper
```
petergphillips, line 16: this was importing the wrong mapper so got an initialisation error

mikehalmamoj (approved):

PR #2947, opened 2026-06-08
SDIT-3771 Temporarily Set email on staff migration if null

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/staff/StaffMigrationService.kt
```diff
@@ -172,7 +172,8 @@ class StaffMigrationService(
 fun StaffDetails.toMigrateStaffRequest(): UserMigrationRequest = UserMigrationRequest(
   user = MigratedUser(
     id = id.toString(),
-    email = email!!,
+    // TODO - ensure the request allows nullable email address
+    email = email ?: "unset",
```
karenmillermoj, line 176: This is just temporary until we get a new version of the API

mikehalmamoj (approved):

PR #3008, opened 2026-06-23
SDIT-3896 Transform sync user to DPS

mikehalmamoj:
Sometimes we end up with the sync user in audit fields that get sync'd to DPS - there are various scenarios where this is valid (mainly for the RaS -> NOMIS -> CAS sync flow, but have seen others too).

DPS have asked that we transform these to their generic "SYS" user which the DPS users understand.


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/MovementHelpers.kt
```diff
@@ -16,3 +16,8 @@ data class ExternalMovementEvent(
   val recordDeleted: Boolean,
   override val auditModuleName: String,
 ) : EventAudited
+
+internal fun String.toDpsUser() = when (this) {
+  "PRISONER_MANAGER_API" -> "SYS"
+  else -> this
+}
```
mikehalmamoj, line 23: This is the transformation - applied it everywhere we send audit fields

steverendell (approved):
lgtm

PR #3040, opened 2026-06-29
SDIT-3845 Take the current term from NOMIS latest booking flag in case an old booking is updated

mikehalmamoj:


karenmillermoj (approved):
lgtm

PR #3043, opened 2026-06-30
SDIT-4001 Temporarily remove downtime in preprod to allow successful transaction reconciliation

karenmillermoj:


helm_deploy/values-preprod.yaml
```diff
@@ -34,7 +34,8 @@ generic-service:
     CONTACT-SENTENCING_COURT-EVENT-UPDATE_IGNORE-MISSING: true
 
   scheduledDowntime:
-    enabled: true
```
karenmillermoj, line 37: I want to see the transactions being added during the overnight batch job, so don't want to just "repair".

steverendell (approved):

PR #3130, opened 2026-07-16
SDIT-4035 Add feature switch to ignore court case linked appearance inserts and updates

mikehalmamoj:
When court scheduler and remand and sentencing sync court case linked court appearances between themselves, we'll need to stop listening for court appearances with court cases in the to-CAS sync. In the meantime we will put this behind a feature switch to support testing.


karenmillermoj (approved):
lgtm

PR #3132, opened 2026-07-16
SDIT-4035 Add feature switch to ignore court case linked appearance deletes

mikehalmamoj:
Similar to #3130 but introduces a feature switch to ignore deletes of court linked court appearances. We need a separate feature switch to inserts and updates because there are different dependencies.

Also fixes a bug where we should have been ignoring deletes that originate in NOMIS, but weren't.


petergphillips (approved):
?LGTM

PR #3155, opened 2026-07-21
SDIT-3764 Pick up latest staff open api docs

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/staff/DummyStaffDpsApi.kt
```diff
@@ -33,15 +34,11 @@ class DummyStaffDpsApi {
       log.info("Migrated staff ${staff.user.staffId}")
     }
 
-  @PutMapping("/prison-users/staff")
+  @PutMapping("/sync/user/{legacyStaffId}")
```
karenmillermoj, line 37: still using dummy endpoints until data corrected

steverendell (approved):
lgtm

PR #3177, opened 2026-07-27
SDIT-4033 Handle RaS linked court movements

mikehalmamoj:
Once court scheduler and remand and sentencing sync between themselves we won't have any mappings for RaS court appearances in the CAS mapping database. Therefore whenever we sync a court movement to CAS we need to send them either the existing CAS ID if there's no court case, or the existing RaS ID if there is, or neither if the court movement is unscheduled.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/court/CourtSchedulerSyncMovementService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/court/CourtSchedulerSyncMovementService.kt
@@ -8,6 +8,7 @@ import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.config.trackEvent
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.courtscheduler.model.CourtEventMovement
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.courtscheduler.model.SyncCourtEventMovement
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.courtscheduler.model.SyncUser
+import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.courtsentencing.CourtSentencingMappingApiService
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.helpers.EventAudited.Companion.EDIT_EXTERNAL_MOVEMENTS_AUDIT_MODULE
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.helpers.TelemetryEnabled
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.helpers.track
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/court/CourtSchedulerSyncMovementService.kt
```diff
@@ -102,14 +105,23 @@ class CourtSchedulerSyncMovementService(
...
-    return nomis.toDpsRequest(dpsCourtAppearanceScheduleId = dpsCourtAppearanceId, dpsCourtMovementId = dpsCourtMovementId)
+  private suspend fun findDpsId(nomisCaseId: Long?, nomisEventId: Long, telemetry: MutableMap<String, Any>): Pair<UUID?, String?> = if (nomisCaseId == null) {
+    tryFetchParent { mappingApi.getCourtScheduleMappingOrNull(nomisEventId)?.dpsCourtAppearanceId }
+      .also { telemetry["dpsCourtAppearanceId"] = it }
+      .let { it to null }
+  } else {
+    telemetry["nomisCourtCaseId"] = nomisCaseId
+    tryFetchParent { sentencingMappingApi.getCourtAppearanceOrNullByNomisId(nomisEventId)?.dpsCourtAppearanceId }
+      .also { telemetry["dpsSentencingCourtAppearanceId"] = it }
+      .let { null to it }
```
mikehalmamoj, line 124: This is the main change - if there's no court case then get the CAS ID from the CAS mappings, otherwise get the RaS ID from the court sentencing mappings.

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/movements/court/CourtSchedulerSyncMovementService.kt
```diff
@@ -102,14 +105,23 @@ class CourtSchedulerSyncMovementService(
...
     val nomis = nomisApi.getCourtMovementOut(prisonerNumber, bookingId, movementSeq)
-    val dpsCourtAppearanceId = nomis.courtScheduleOutId
+    val (dpsCourtAppearanceId, dpsSentencingCourtAppearanceId) = nomis.courtScheduleOutId
       ?.also { telemetry["nomisEventId"] = it }
-      ?.let {
-        tryFetchParent { mappingApi.getCourtScheduleMappingOrNull(it)?.dpsCourtAppearanceId }
-          .also { telemetry["dpsCourtAppearanceId"] = it }
-      }
+      ?.let { findDpsId(nomis.caseId, nomis.courtScheduleOutId, telemetry) }
+      ?: (null to null)
```
mikehalmamoj, line 111: And here is where the CAS and RaS mappings are used. Note that if there is no court schedule ID then both are null - this is an unscheduled court movement.

andymarke (approved):

PR #3178, opened 2026-07-27
SDIT-4058 updating the api docs from the mapping service. Only raise …

richardpopple:
…a religion create event if there is no top level mapping for the offender


petergphillips (approved):
LGTM

PR #3180, opened 2026-07-28
SDIT-4058 spliting up the update event to determine if this is a simp…

richardpopple:
…le text change or a new belief

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/coreperson/CorePersonSynchronisationBeliefsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/coreperson/CorePersonSynchronisationBeliefsService.kt
@@ -23,7 +23,6 @@ import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.nomisprisoner.mod
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.InternalMessage
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.SynchronisationQueueService
 import uk.gov.justice.digital.hmpps.prisonerfromnomismigration.service.SynchronisationType
-import kotlin.collections.plus
 
 private const val TELEMETRY_PREFIX = "coreperson-beliefs-synchronisation"
 typealias Telemetry = MutableMap<String, Any>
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/coreperson/CorePersonSynchronisationBeliefsService.kt
```diff
@@ -39,66 +40,77 @@ class CorePersonSynchronisationBeliefsService(
...
-            }
-          }
-        }
+        // There is no mapping so we cannot ignore this as there will be no update because no beliefs exist yet.
+        createBelief(telemetry, getNomisOffenderBelief(event), offenderIdDisplay, event.offenderBeliefId)
       }
     }
   }
 
   suspend fun offenderBeliefUpdated(event: OffenderBeliefEvent) {
```
richardpopple, line 64: removing the telemetry name is this method can now do an update or an create depending on what has actually happened to generate the update event

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/coreperson/CorePersonSynchronisationBeliefsService.kt
```diff
@@ -39,66 +40,77 @@ class CorePersonSynchronisationBeliefsService(
...
-              corePersonCprApiService.syncUpdateOffenderBelief(
-                offenderIdDisplay,
-                mapping.cprId,
-                this,
-              )
+      val allBeliefs = corePersonNomisApiService.getOffenderReligions(event.offenderIdDisplay)
+      val currentBelief = allBeliefs.first()
+      val currentBeliefMapping = religionsMappingService.getReligionByNomisIdOrNull(currentBelief.beliefId)
+      if (currentBeliefMapping != null) {
+        // This event can only be a simple update of the comments field.
```
richardpopple, line 77: Let the code pretty much as is for the simple update case. Once CPR have accepted the PR changing the update request we can probably simplify this - however we will still need to do additional look ups as the belief in question is not necessarily the current belief we have looked up

src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/coreperson/CorePersonSynchronisationBeliefsService.kt
```diff
@@ -39,66 +40,77 @@ class CorePersonSynchronisationBeliefsService(
...
+                  mapping.cprId,
+                  this,
+                )
+              }
             }
-          }
+        }
+      } else {
+        // This event indicates a new active belief has been created.
+        telemetry += ("nomisId" to currentBelief.beliefId)
```
richardpopple, line 93: I think this is the correct behaviour? The update event would not have come from the event that has been created. But I think that the nomisId here should be from the event that been created?

petergphillips (approved):
LGTM

PR #3186, opened 2026-07-28
SDIT-4079 Allow transaction updated events to behave as insert events if no dps mapping

karenmillermoj:


src/main/kotlin/uk/gov/justice/digital/hmpps/prisonerfromnomismigration/finance/TransactionSynchronisationService.kt
```diff
@@ -76,15 +75,12 @@ class TransactionSynchronisationService(
       telemetryClient.trackEvent("transactions-synchronisation-updated-skipped", event.toTelemetryProperties())
       return
     }
-
+    // It is safe to sync an update which has no Dps mapping (it could be a historical update i.e. insert occurred before sync to dps began)
```
karenmillermoj, line 78: added a big comment as this is different behaviour to a normal update

petergphillips (approved):
LGTM

PR #3197, opened 2026-07-29
SDIT-4033 Fix flaky tests to account for retries

mikehalmamoj:
Some tests rely on a retry but we were jumping to the next test before the retry had happened - so we've seen some bleeding into other tests on CI. Fix is to wait for the retry.


karenmillermoj (approved):
lgtm

